PR #67, opened 2013-11-17
deref a version set as *VERSION = \...

charsbar:
$v here in line 460 should be a scalar reference when the left part is *VERSION, and should be dereferenced before it is passed to the parent. An example to test is Acme::Lvalue 0.03 (which is still on CPAN).


rjbs:
@charsbar I have added a test for this case, and it makes the patch seem unneeded: e42cc86

charsbar:
@rjbs awesome! Thank you.

PR #68, opened 2013-11-17
undefined provides package version should always be 'undef' no matter if it exists or not

charsbar:
Sometimes "provides" package version may become explicit undef, which passes the "exists" test in line 259 and causes an unini warning in line 261.


rjbs:
Replaced by https://github.com/andk/pause/pull/330

PR #70, opened 2013-11-17
VERSION comparison shouldn't be detected as VERSION definition

charsbar:
An example to test is https://metacpan.org/source/DOY/KiokuDB-0.54/lib/KiokuDB/Util.pm#L179 . This kind of version comparison shouldn't be detected as VERSION definition.


rjbs:
Replaced by https://github.com/andk/pause/pull/331

PR #71, opened 2013-11-17
skip if provides file field is empty

charsbar:
An example to test is https://metacpan.org/source/RIBASUSHI/DBIx-Class-Manual-SQLHackers-1.3/META.yml#L23 . Empty file field causes several troubles such as bogus stats etc.

### DIFF
--- a/lib/PAUSE/dist.pm
+++ b/lib/PAUSE/dist.pm
@@ -787,6 +787,8 @@ sub examine_pms {
     }
   } elsif (2==$indexingrule) { # a yaml with provides
     while (my($k,$v) = each %$provides) {
+      next if ref $v ne ref {};
+      next if !defined $v->{file} or $v->{file} eq '';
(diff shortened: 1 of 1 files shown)


rjbs:
@charsbar I have re-indexed DBIx-Class-Manual-SQLHackers-1.3.tar.gz and don't see a problem.  There *is* a non-compliant no_index, but there are *also* no `.pm` files in the distribution (outside of `inc`), so I wonder if the problem was something else.

That said, I will write a test case for something that has `.pm` files *and* a bogus no_index!

rjbs:
Oops, I meant `provides`, and not `no_index`. :)

rjbs:
I am closing this MR in favor of https://github.com/andk/pause/pull/329 which is more modern, rebased, and has a test -- but there was definitely some weirdness here, thanks!

PR #133, opened 2014-11-19
Permit pumpking

rjbs:
We discussed this in email: a pumpking-bit holder should be able to upload a new maint perl, and its single-life modules should be indexed.  For example, `Benchmark`  should point to 5.20.1, not 5.20.1, right now.  Andreas said, "I could've sworn the code worked this way already!"

Steve Hay found the old indexer report and it made things clear:  the "must have rights to package reflecting module name" was getting in the way!  I've special cased the check it uses:  when it wonders "does the uploader have rights to the package `perl`, it checks the pumpking bit rather than the module permissions list.

Once this is merged, we should re-re-index perl-5.20.1!

### DIFF
--- a/lib/PAUSE.pm
+++ b/lib/PAUSE.pm
@@ -496,6 +496,29 @@ sub _now_string {
   return $self->_time_string(time);
 }
 
+sub user_has_pumpking_bit {
+  my ($self, $user) = @_;
+
+  use DBI;
+  my $adbh = DBI->connect(
+    $PAUSE::Config->{AUTHEN_DATA_SOURCE_NAME},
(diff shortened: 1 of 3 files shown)


karenetheridge:
> For example, Benchmark should point to 5.20.1, not 5.20.1

Is this meant to read "should point to 5.20.1, not 5.20.0"?

Note there are also lots of modules presently indexed to perl-5.18.4.

rjbs:
Yes.  There is, I *think* a separate issue for the fact that later uploads trump later version for perl core.

andk:
Merged April 2015. Closing now. Thanks!

PR #134, opened 2014-11-21
Give full indexing report even if dist name perm fails

dagolden:
This commit makes a distribution name permission failure merely add an
informative header to the message, but otherwise allows the ordinary
full content of the report.  However, any "indexed" modules are fixed up
to show that they are not indexed.

This should allow better diagnostic information when dist name problems
occur.


karenetheridge:
+1

andk:
merged as de453e1 and deployed (during #perlqah2015)

karenetheridge:
danke danke!

PR #137, opened 2014-11-30
Document classes, and current checks on distros before indexing

mohawk2:


andk:

PR #147, opened 2015-03-17
HTML entity-escape arbitrary data being put into HTML

rjbs:
(subject and patch speaks for itself)


andk:
Thanks, this was pulled with c38ccbc!

PR #149, opened 2015-04-01
More speed fixes for peek_perms and share_perms

ilmari:
Restore the performance win from commit 2fe5d79 that was accidentally broken in commit 3c90b40, and apply a similar fix to `share_perms`.

This speeds up getting the list of modules to relinquish co-maint from from ~2m45s to ~0.1s on my laptop.


karenetheridge:
+:100:

andk:
+:THANKS:

PR #152, opened 2015-04-16
start using bcrypt for PAUSE passwords

rjbs:
DES stinks.  Let's use Bcrypt!

**Achtung!**  We have to have more space for hashes in the database with this change!  The password field should store 60 characters.  (I see it uses char rather than varchar.  I think you will want to switch to varchar, unless you've got something trimming whitespace from the end.  I'm not sure.  I haven't used the char type in a decade.)

Other than that, I believe this PR is ready to go.

Note the new `maybe_upgrade_stored_hash` function in PAUSE::Crypt.  Once we enable that, it will start upgrading users to new passwords.

Let me know when this is live, and I will convert my account ASAP! 😄


dagolden:
FWIW, here's the article I remembered that talked about work factors.  Their "starting point" for work factor was 11, so I think 12 is a good place to be, given that a slight delay logging into PAUSE is pretty immaterial.

http://chargen.matasano.com/chargen/2015/3/26/enough-with-the-salts-updates-on-secure-password-schemes.html

dougwilson:
:+1:

PR #153, opened 2015-04-16
CGI.pm dist name is now just CGI

rjbs:
Once, CGI.pm was in a dist called `CGI.pm-4.56.tar.gz`.  No more!

This special case can go away.


andk:
Merged in 25f18f43

PR #156, opened 2015-04-17
enable auto-rehashing of passwords

rjbs:


andk:
Thanks, applied + deployed

PR #157, opened 2015-04-17
remove embarrassing syntax error

rjbs:


andk:
Thanks applied+deployed in 71cca7d

PR #160, opened 2015-04-18
index single-life dev versions in core

rjbs:
If the pmfile seems to be a dev version, we skip it... but not if it's in perl.  Historical example:  perl-5.18.0 contains POSIX 1.32, but perl-5.20.0 contains 1.38_03.  We would then skip the new one, leaving some packages pointing at an older version of perl.

We don't need to account for single- versus dual-life, because the check for dual-life packages still applies elsewhere.

### DIFF
--- a/cron/mldistwatch
+++ b/cron/mldistwatch
@@ -1,4 +1,4 @@
-#!/usr/local/bin/perl -w
+#!/usr/local/bin/perl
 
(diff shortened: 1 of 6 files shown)


andk:
Thanks, applied as 06c1eec and deployed and perl-5.20.2 again scheduled for indexing

PR #165, opened 2015-04-22
the build-fresh-cpan tool

rjbs:
one-off-utils/build-fresh-cpan -- this will keep getting better, but it's a tool to start a new "CPAN" and upload a bunch of tarballs; then you can see how they'd get indexed.  It's good for figuring out what the heck crazy thing happened with a tarball you've seen badly indexed.


andk:
Merged as e4c03bf8fe07dc57b72f095329b768a4602df8c3 and deployed. Thanks!

PR #172, opened 2015-05-08
cope with fatal errors in manifind

rjbs:
manifind runs File::Find, which might die on recursive
symlinks; when that happens, PAUSE should not die


karenetheridge:
manifind probably shouldn't die either. ExtUtils::Manifest iz dum. :(

andk:

PR #173, opened 2015-05-08
be stricter in when finding packages in files

rjbs:
This will prevent us from finding a package in this line of code:

  my $x = "package Foo;";

This manifested in the wild in PRBRENAN/Math-Algebra-Symbols-1.21.tar.gz

### DIFF
--- a/lib/PAUSE/pmfile.pm
+++ b/lib/PAUSE/pmfile.pm
@@ -225,15 +225,16 @@ sub packages_per_pmfile {
 
         if (
             $pline =~ m{
-                      (.*)
-                      (?<![*\$\\@%&]) # no sigils
-                      \bpackage\s+
+                      ^
+                      [\s\{;]*
+                      package
+                      \s+
                       ([\w\:\']+)
                       \s*
                       (?: $ | [\}\;] | \{ | \s+($version::STRICT) )
                     }x) {
-            $pkg = $2;
-            $strict_version = $3;
+            $pkg = $1;
+            $strict_version = $2;
             if ($pkg eq "DB"){
                 # XXX if pumpkin and perl make him comaintainer! I
                 # think I always made the pumpkins comaint on DB


rjbs:
We discussed this in Berlin, and the concern is that it cause changes to old behavior.  I began a process of re-indexing all of CPAN, but it was clearly going to take a CPU-week.

My next idea is to search for distributions with lines matching one regex and not the other, then inspect those specifically.

rjbs:
I've got a job running now, but I need to go to bed.  In the morning I'll put the results in a commit on this branch.  They are quite interesting so far!

rjbs:
So!  I pushed [the report file](https://github.com/rjbs/pause/blob/4dae699df7b89d930c28c21822244a56869672ee/tmp/package-change-report.txt) after making a few changes after the first run.

The fields are:  filename, line number, matched by old regex?, matched by new regex?  I didn't log the dist file, oops!  I think we're okay without it for now.

The great majority of changes seem to be packages in strings, sometimes being eval'd.  Plenty of them, though, are just lines containing things like `$name = "the package Foo";`

Other interesting constructs that this change would *stop* indexing include:

* `map {; package XYZ; ... }`
* `do { package XYZ; ... `
* `my $code = sub { package XYZ; ... };`

I think this change is worth making, but I cede that it is at least not clear.

karenetheridge:
All of these code snippets need to be added to Module::Metadata's test corpus. As far as I know it's still in our long term plans to cut PAUSE over to using Module::Metadata.

(@rjbs: please feed me a list when available and I'll add them, adding TODO markers as needed.)

rjbs:
Is the list in the "report file" linked to above sufficient?  (If not, why not?)

rjbs:
I'm removing the report file from the PR and attaching it here.

rjbs:
[package-change-report.txt](https://github.com/andk/pause/files/11352828/package-change-report.txt)

charsbar:
cf. https://github.com/charsbar/Parse-PMFile/commit/53729d72eae3015298a858d46afda3529d40881c

andk:
Thanks, applied, and deployed just now.

PR #184, opened 2015-09-08
Fix regex for META6.json detection.

jdv:


karenetheridge:
I agree with @rjbs on irc -- let's instead fix the tools that are putting META6.json in a subdirectory.

rjbs:
It's not quite that. META6.json is *always* in a subdirectory in a well-formed dist.  If you upload `Awesome-Dist-1.23.tar.gz` then all the files should be in `./Awesome-Dist-1.23` and the regex did not allow that.

On the other hand, the change that gets made should not allow the file to live just anywhere, but it should be in the "main" directory in the dist archive.  My suggestion was to add an assertion that there *is* only a top-level directory.  This would reject tarbomb archives, which I *believe* are now carefully handled elsewhere.

PR #186, opened 2015-09-14
Account for $/ being "" around these parts.

jdv:


andk:
Thanks!

Deployed at 2015-09-16 03:41 UTC

PR #190, opened 2015-11-04
Add link to CPAN Testers matrix in the first email from PAUSE after an upload

neilb:
Hi Andreas,

This is something I suggested a while back: give nudge people to go look at CPAN Testers
after uploading to PAUSE.

If CPAN::DistnameInfo can work out a dist name from the release file name,
then we add some simple test to the email, giving them the URL for
the CPAN Testers matrix.

Cheers,
Neil


andk:

Issue #191, opened 2015-12-20
version index limited to 16 characters, indexing messes up silently.


kentfredric:
As exemplified by Net::FullAuto, as soon as you exceed 16 characters, new releases will start getting indexed at a truncated version.

This is also imagined to happen with very large integer values, which might result in strange behaviour as  "1234" becomes "123" ( except with 16 instead of 3 )
```
 $ wget -qO- cpan.metacpan.org/modules/02packages.details.txt.gz | zgrep Net::FullAuto
Net::FullAuto                  0.99999999999999  R/RE/REEDFISH/Net-FullAuto-0.9999999999999905.tar.gz
Net::FullAuto::FA_Core            undef  R/RE/REEDFISH/Net-FullAuto-0.9999999999999905.tar.gz
Net::FullAuto::FA_DB              undef  R/RE/REEDFISH/Net-FullAuto-0.9999999999999905.tar.gz
Net::FullAuto::Getline            undef  R/RE/REEDFISH/Net-FullAuto-0.9999999999999905.tar.gz
Net::FullAuto::MemoryHandle       undef  R/RE/REEDFISH/Net-FullAuto-0.9999999999999905.tar.gz
```

And the last 5 releases have been similarly indexed, and indexed with the truncated version:
```
 wget -qO- http://cpanmetadb.plackperl.org/v1.0/history/Net::FullAuto | grep '0.99999999999999'
Net::FullAuto                  0.99999999999999  R/RE/REEDFISH/Net-FullAuto-0.99999999999999.tar.gz
Net::FullAuto                  0.99999999999999  R/RE/REEDFISH/Net-FullAuto-0.9999999999999901.tar.gz
Net::FullAuto                  0.99999999999999  R/RE/REEDFISH/Net-FullAuto-0.9999999999999902.tar.gz
Net::FullAuto                  0.99999999999999  R/RE/REEDFISH/Net-FullAuto-0.9999999999999903.tar.gz
Net::FullAuto                  0.99999999999999  R/RE/REEDFISH/Net-FullAuto-0.9999999999999904.tar.gz
```

karenetheridge:
I don't really see a problem with this, so long as the PAUSE index email receipt indicates what the indexed version was.  After all, we have to have a limit *at some point* if we expect versions to be evaluated in numeric context.

Especially given that Net::FullAuto is the first and perhaps only distribution to ever encounter this situation, I'd classify it as "not a bug -- works as intended".

kentfredric:
[...]

PR #195, opened 2015-12-23
reject overlong versions (and other unstorable data)

rjbs:
This PR has two parts:

1. it rejects versions over 16b long with an explanation in the indexer report; tests included
2. it puts MySQL into strict mode to prevent silent truncation of other data

In the past, if a field was too long to be stored, it was silently truncated.  This is what led to the problem described in #191, where long versions were truncated but the contents still indexed.  With change 2 above, this will result in a MySQL insertion error.  We can then add application-level checking and error reporting as needed.  I don't expect this to be a big problem.

Note that because the test suite runs against SQLite, I did not test the strict mode change.  I expect it to be no problem, but proceed with a bit of caution.

I believe the long version change is quite safe.


lib/PAUSE/mldistwatch/Constants.pm
```diff
@@ -21,6 +22,7 @@ use constant OK => 1;
 
 our $heading = {
   EBADVERSION() => "Version string is not a valid 'lax version' string",
+  ELONGVERSION() => "Version string is exceeds maximum allowed length of 16b",
```
zoffixznet: There's a typo here: `string is exceeds` should be `string exceeds`

t/mldistwatch-misc.t
```diff
@@ -252,6 +252,39 @@ subtest "check various forms of version" => sub {
...
+  $pause->file_not_updated_ok(
+    $result->tmpdir
+           ->file(qw(cpan modules 02packages.details.txt.gz)),
+    "there were no things to update",
+  );
+
+  my $etoolong = sub {
+    like(
+      $_[0]{email}->get_body,
+      qr/Version string is exceeds maximum allowed length/,
```
zoffixznet: The typo I mentioned above is also present in the test here. `string is exceeds` should be `string exceeds`

andk:
Rebased, merged, pushed, deployed to pause. Thanks!

PR #200, opened 2016-01-28
remove link to non-existent pause99_edit_mod_2

karenetheridge:
Remaining 'm' entries in the permissions database must be edited by hand, as
all modulelist functionality has been removed from the website.

In the future, we should bulk-change all 'm' entries to 'f' so we can remove
mention of the modulelist altogether.


andk:
Thanks, merged and deployed to pause

Issue #202, opened 2016-03-10
"distribution not indexed" error message is not always correct


karenetheridge:
See the message to module-authors: http://www.nntp.perl.org/group/perl.module-authors/2016/03/msg10720.html

```
The following report has been written by the PAUSE namespace indexer.
Please contact modules@perl.org if there are any open questions.

  User: GUIMARD (Xavier Guimard)
  Distribution file: Apache-Session-Browseable-1.2.tar.gz
  Number of files: 41
  *.pm files: 23
  README: Apache-Session-Browseable-1.2/README
  META-File: Apache-Session-Browseable-1.2/META.json
  META-Parser: Parse::CPAN::Meta 1.4414
  META-driven index: yes
  Timestamp of file: Wed Mar  9 05:31:13 2016 UTC
  Time of this run: Wed Mar  9 05:33:05 2016 UTC

This distribution name can only be used by users with permission for the
package Apache::Session::Browseable, which you do not have. No modules
will be indexed.
```

In fact, the user does have permission for Apache::Session::Browseable. The actual problem was that there was no lib/Apache/Session/Browseable.pm (with appropriate `package` statement) inside the distribution, so the error message was highly confusing to the author.

wolfsage:
I'll try to hammer this one out today

wolfsage:
PR attached.

Strangely enough though, I don't believe the user should have gotten an error at all when uploading.

You don't need a package called <distname>.pm if you already have permissions in 06perms.txt to a module called <distname>. When this new restriction was put in place, 06perms.txt was backfilled to be correct for modules like this one, so there shouldn't have been a problem.

I'm not entirely sure *what* happened here.

wolfsage:
Ah, also see PR #168 for an alternative patch

PR #207, opened 2016-04-21
Add note that upload may have failed because a package is missing.

wolfsage:
For GH #202.

Adds a hopefully helpful message if someone gets a permission failure when uploading a dist.

The message is:

    You appear to be missing a .pm file containing a package matching the dist name ($pkg). Adding 
    this may solve your issue.

This message may show up in some situations where it is a lie (like uploading fails because you don't actually own the dist in question).

If we want to fix that, please let me know.


andk:

PR #208, opened 2016-04-21
Topic/keep emails in build fresh cpan

wolfsage:
This allows closer examination of the email messages that pause is generating when using build-fresh-cpan.

CC @rjbs


rjbs:
This looks good to me.

andk:
Merged. Thanks!

Issue #119, opened 2014-05-27
include sha, not just md5, in "has entered cpan" emails


rjbs:
We include this:

>   md5: 20cbecd4e9e880ee7a50a136c8b1484e                                               

...but "new perl" announcements give SHA1 sum.  We should include that in the email.

wolfsage:
Possible patch attached

charsbar:
This seems to have been fixed already in 2016.

PR #211, opened 2016-04-22
Add sha1 checksums along side md5 checksums.

wolfsage:
For GH #119.

The bin/paused portion is untested.

The lib/PAUSE.pm portion changes output that mirrormail.pl generates, so if anything is parsing that that may be a problem?


andk:

PR #214, opened 2016-04-23
set $PAUSE::Config->{SERVER_NAME} to uri->host if it is set

### LINKED ISSUE #213: pause.cpan.org now has pause.perl.orgs cert (cert fail)
I tried to log into pause.cpan.org and got a certificate warning due to it using the pause.perl.org certificate. This issue is new as of this evening.

charsbar:
... so that even when someone goes to http://pause.cpan.org, (s)he will go to https://pause.perl.org next. This should fix #213


exodist:
can confirm, is fixed, thanks!

PR #215, opened 2016-04-23
made /incoming and /pub/PAUSE alive again

charsbar:
I added two more mounts I've overlooked


andk:
Merged and deployed a few minutes ago, Thanks.

PR #217, opened 2016-04-23
Various web tests

pjlsergeant:
* Adds a small framework to make adding webtests easier
* Tests:
  * basic authentication
  * changing password
  * email_for_admin
  * various file listing

@charsbar

### DIFF
--- a/t/lib/pause_1999/Test/SiteModel/Parser.pm
+++ b/t/lib/pause_1999/Test/SiteModel/Parser.pm
@@ -66,7 +66,7 @@ our %parsers = (
     },
     yaml => sub {
         my ( $tree, $content ) = @_;
-        return Load $content;
+        return scalar Load $content;
     },
 );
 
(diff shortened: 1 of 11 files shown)


t/lib/pause_1999/Test/SiteModel/Parser.pm
```diff
@@ -0,0 +1,109 @@
...
+            }
+            else {
+                ();
+            }
+        } split( m!<br />|\n!, $pre->as_HTML );
+        return \@files;
+    },
+    yaml => sub {
+        my ( $tree, $content ) = @_;
+        return Load $content;
```
rjbs: In list context, `Load` returns a list.  Although we can take it as *really* unlikely that we ever have more than one document in `$content`, but still, we should put this Load into scalar context with `scalar`, lest we screw up a hash horribly.

pjlsergeant:
Incorporated change from @rjbs comments

andk:
Merged (in two chunks) and deployed a minute ago. Thanks!

PR #218, opened 2016-04-23
Divert various warn() messages to go through Log::Dispatch

pjlsergeant:
Creates `main::warn()`, which routes through `Log::Dispatch` as per @charsbar 

``` perl
sub warn {
    my ($self, $message) = @_;
    local $Log::Dispatch::Config::CallerDepth += 2;
    $self->{REQ}->logger->({level => 'error', message => $message });
}
```

Then rewrites the various warns to use that instead.


rjbs:
pause_1999 has been superseded by pause_2017, and we have agreed to close this request unapplied.  Thanks for the work, and we apologize that it wasn't applied earlier!

PR #221, opened 2016-04-24
First stab at a travis job

pjlsergeant:
Builds PAUSE on Travis. Output will be considerably less noisy when #218 has been applied, but works and is useful now.


pjlsergeant:
https://travis-ci.org/pjlsergeant/pause

PR #222, opened 2016-04-24
Review/new default permissions

colinnewell:
These are the changes to the default permissions for new modules added to a distribution.  The code for x_authority has not been removed, but the behaviour has changed to the new copy all comaintainers behaviour.  I could probably do something to retain that behaviour, but I'm not sure if that's really desirable.

I'm wondering if I need to test any more scenarios.

/cc @rjbs


rjbs:
Thanks!  I will try to review this in the next day or two.

rjbs:
> > rjbs commented 27 days ago
> Thanks! I will try to review this in the next day or two.

I'm the worst.  Reviewing now.

lib/PAUSE/package.pm
```diff
@@ -83,12 +83,27 @@ sub give_regdowner_perms {
...
-  if ($sth_mods) { # make sure we regard the owner as the owner
+                                WHERE  modid = ?
+                                ";
+  my @args = ($package);
+  unless(lc($main_package) eq lc($package)) {
+    $sql .= q{
+    UNION
+    SELECT userid
+    FROM   perms
+    WHERE  LOWER(package) = LOWER(?)
```
rjbs: Not a big deal, but: this is MySQL, so this `=` is already case-insensitive, right?  If we're going to use an explicit LOWER on both sides (which seems fine) we should probably also add it to the `WHERE modid = ?` above, too.

karenetheridge: Case sensitivity is controlled by collation settings, which can be set globally, at the database level and at the table level. The default *might* be case-insensitive, but one would need access to the database to be sure.

colinnewell: I tried to be consistent with the existing code with regards to the case sensitivity.  Plus don't the tests use SQLite so it needs to work in a DB server independent manner?

No lc on the modid was a case of not touching existing code any more than I needed to.  I didn't even consider touching that query.

I rather suspect that that whole bit will be chopped out as I think it's related to that 03modules thing that's effectively been removed.  Touching it further is probably a waste of time.

rjbs:
I think this looks good!

It took me a bit longer to remember just what every method was meant to do, but once I did, I believe it is correct.  I think we should probably merge this and add an item to the PAUSE news.

andk:
Thanks, applied and deployed to live just now. paused restarted.

PR #227, opened 2016-07-19
don't index packages seen in example/demo directories

neilb:
Hi Andreas,

This change means that PAUSE won't look in examples/, example/, eg/, or demo/ when looking for package statements.

This will help resolve a number of permissions conflicts, where people have demo classes called things like `testclass.pm`, `example.pm`, etc.

Neil


neilb:
Going to change this.

dolmen:
Ideally this list of files/paths should be available as a module to help to build developer tools. For example a dzil plugin that, on release, reports files that will be indexed.

PR #228, opened 2016-07-21
don't index packages seen in example and demo directories

neilb:
This skips indexing of packages seen in the following directories:

 * examples
 * example
 * ex
 * eg
 * demo

These all have enough usage in the wild to make it worth adding them to the skip list.

Skipping these will help me resolve a number of PAUSE permission conflicts.


neilb:
\o/

Can you let me know when this goes live please, as I'll then be able to work on a number of permissions conflicts :-)

Cheers,
Neil

andk:
Sorry, forgot to add this to the pr. It went live just a few moments after the merge, i.e.
Sun Jul 24 23:15:41 2016 +0200

Issue #232, opened 2017-02-10
PAUSE will give first-come if you upload a package with a case-varied name of an existing first-come package


neilb:
OALDERS (@oalders) has first-come on Acme::Odometer:

    Acme::Odometer,OALDERS,f

I just uploaded Acme-Playpen dist, with Acme::odometer (different only in the case of the first 'o'). Mine was a cuckoo package, but that doesn't make any difference (tested with another upload).
And I was granted first-come:

    Acme::odometer,NEILB,f

This is related to [#175](https://github.com/andk/pause/issues/175), but that issue (also submitted by me!) seems to suggest that in the past I wouldn't have been granted first-come!

Neil

neilb:
This was fixed during the Toolchain Summit. I did another release this week to check, and can confirm that it is fixed.

PR #233, opened 2017-03-01
Check perms on package name case insensitively

neilb:
Right now, if someone else already has first-come on Foo::Bar,
then I can upload Foo::BAR, and as long as my version numbeer, then I'll get first-come on Foo::BAR,
permissions on both Foo::Bar and Foo::BAR will be listed in 06perms.txt,
but my module will be indexed.

This change checks permissions case insensitively, so I can't get perms on any case-variant of Foo::Bar

This addresses [issue 232](https://github.com/andk/pause/issues/232).

Cheers,
Neil

### DIFF
--- a/lib/PAUSE/package.pm
+++ b/lib/PAUSE/package.pm
@@ -187,10 +187,13 @@ sub perm_check {
       return 1;           # P2.1, P3.0
   }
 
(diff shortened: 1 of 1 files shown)


neilb:
This wasn't the right fix, and the bug has been fixed correctly now :-)

PR #234, opened 2017-05-11
Exercises a bug where comaint is erroneously awarded to submodule in a package

pjlsergeant:
This test fails only intermittently, as the error itself occurs only intermittently. This is less than ideal, but hard to think of a sensible workaround


haarg:
I believe this was caused by fff5aa175d821ac04ed7fdd3ffee3f4eb914a06d.

andk:
Tentative fix pushed to master branch. More eyeballs welcome.

andk:
Explicitly pinging @colinnewell and @rjbs. Maybe this change breaks what you intented to achieve, but at least it does not break your tests. Please be prepared during your next uploads. It's not yet deployed but will be RSN unless somebody raises objections.

colinnewell:
I'm reasonably confident that if my tests pass then the core of what I wanted to achieve is still occurring.  All we wanted to do was ensure that new modules that were indexed had consistent permissions.  Excluding modules that shouldn't be indexed is a whole different kettle of fish.

I would suggest adding a test that double checks that if OOOPPP had comaint on that existing module that the module does get indexed correctly.

andk:
Thanks for your quick response, Colin. I've added two lines to the test that may or may not represent your idea. All tests still pass.

It's too late to deploy today, I'll give it a go tomorrow morning European time if no showstopper is found until then.

colinnewell:
Apologies for introducing that bug.

PR #238, opened 2017-05-12
Added an test to check that with comaint we can upload the module.

colinnewell:
The permissions end up a little odd, but we can upload a module
that subsumes a previous one if we have comaint, and we don't end
up grabbing first come as part of that process.

This was the test I was suggesting for #234.


andk:
Thanks a lot, Colin, highly appreciated! Pulled in and about to be deployed.

karenetheridge:
@colinnewell++

Issue #187, opened 2015-09-14
Web interface reports different results from 06perms.txt for package Alien


neilb:
Last week I asked you (ANDK) to change `Alien`'s permission for `ABERGMAN` from **m** to **f**, which you did. I then transferred ownership to `PLICEASE`. When I looked at `06perms.txt` there was no entry for ABERGMAN, as he asked me to drop his co-maint as well, which I had done.

But then PLICEASE mentioned that the web interface was telling him that ABERGMAN still had module-list permission ('m'). Here's what the web interface tells me right now:

>  ![screen shot 2015-09-14 at 22 53 45](https://cloud.githubusercontent.com/assets/478461/9862856/8027c612-5b33-11e5-8dd1-86146894d101.png)

Which is slightly odd, since it's saying ABERGMAN has modulelist, but on the other rows it's saying PLICEASE has ownership!

I just grabbed the latest 06perms.txt, and ran:

```
% grep '^Alien,' 06perms.txt
Alien,ABERGMAN,c
Alien,ETJ,c
Alien,JBERGER,c
Alien,NEILB,c
Alien,PLICEASE,f
Alien,ZMUGHAL,c
```

(I asked PLICEASE to try giving me co-maint, to test if he had ownership, which is why you see me having co-maint).

So 06perms is saying that ABERGMAN has co-maint but the web interface is saying he has 'm', and earlier today 06perms wasn't listing ABERGMAN at all for `Alien`, which is what I expected.

Cheers,
Neil

andk:
Confirmed, the web interface is wrong, 06perms is right

andk:
Thanks, charsbar! Deployed.

PR #240, opened 2017-05-14
don't peek mod table to find who have permisson for a module

charsbar:
This should address #187.


charsbar:
Merged with Andreas' consent.

PR #241, opened 2017-05-14
remove more mmods stuff related to #240/#187

charsbar:
This should fix a similar issue as #187. In this case, the current code removes some modules someone has comaintainership for. For example, ABERGMAN has a comaintainer bit for Alien module, but Alien is not listed in his "change permissions" page's "Give up co-maintainership" list. The main reason of this issue was all_mmods doesn't see mlstatus column and still takes "deleted" entries into account. If you prefer a smaller fix, I'll discard this PR and tweak the problematic SQL instead.


andk:
Applied as abb26661, then added a tweak to an older buglet in the neighborhood of this code as d6408893a. Thanks!

PR #242, opened 2017-05-15
Update source-url in Perl 6 META6.json files when extracting

niner:
The source-url field in the META6.json is used by installers to actually fetch the dist. Without this source-url, the installer has to guess the filename extension of the archive (which is what zef currently does). I fully expect installers to actually replace the first part of the URL with the base URL of a CPAN mirror. This has been agreed to by zef's author.

Please note that while I've tested the code in a separate test script, I have not tested it as part of PAUSE itself. It works in the test script and does compile in PAUSE, but it's possible that I have misunderstood what the contents of some of the variables or configuration values should be.

The server load should hardly be affected by this patch as the change only affects uploads of Perl 6 modules and anyway JSON::XS is very fast.

### DIFF
--- a/lib/PAUSE/dist.pm
+++ b/lib/PAUSE/dist.pm
@@ -9,6 +9,7 @@ use List::MoreUtils ();
 use PAUSE ();
 use Parse::CPAN::Meta;
 use PAUSE::mldistwatch::Constants;
+use JSON::XS ();
 
 # ISA_REGULAR_PERL means a perl release for public consumption
 # (and must exclude developer releases like 5.9.4). I need to
@@ -1045,7 +1046,11 @@ sub extract_readme_and_meta {
   for my $metafile ($json || $yaml) {
     if (-s $metafile) {
       $self->{METAFILE} = $metafile;
-      File::Copy::copy $metafile, "$MLROOT/$sans.meta";
+      if ($self->perl_major_version == 6) {
+        $self->write_updated_meta6_json($metafile, $MLROOT, $dist, $sans);
+      } else {
+        File::Copy::copy $metafile, "$MLROOT/$sans.meta";
+      }
       utime((stat $metafile)[8,9], "$MLROOT/$sans.meta");
       PAUSE::newfile_hook("$MLROOT/$sans.meta");
       my $ok = eval {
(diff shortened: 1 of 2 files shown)


lib/PAUSE/dist.pm
```diff
@@ -9,6 +9,7 @@ use List::MoreUtils ();
 use PAUSE ();
 use Parse::CPAN::Meta;
 use PAUSE::mldistwatch::Constants;
+use JSON::XS ();
```
karenetheridge: We should not use JSON::XS directly in toolchain code -- we already have a wrapper around various JSON backends (that the user can switch between through environment variables) in Parse::CPAN::Meta::load_json_string.

charsbar: PAUSE indexer is a standalone web service which is not expected to be run by others, and thus it doesn't need a wrapper at all. JSON::XS is the one that has been used by PAUSE for a long time, and it is already listed in PREREQ_PM in Makefile.PL. This usage is completely right.

jdv:
Wouldn't a simpler way forward be to just validate the entire dist, including the source-url
meta field, and kick the upload as a way to keep things cleaner?  I believe that's more the
way its done on the p5 side.  That would allow for more checks to be added in the future
and would not hinder things like dist signing.

niner:
This is not about validation but about pointing the installer to the right tar ball. The ecosystem list indexes the extracted .meta files and thus gets the source-urls.
Updating the source-url in the extracted .meta file does not hinder signing as the tar ball remains completely untouched.

andk:
Thanks, applied and deployed on PAUSE just now

lib/PAUSE/dist.pm
```diff
@@ -1063,6 +1068,32 @@ sub extract_readme_and_meta {
...
+    $json->decode(join '', <$meta_fh>);
+  };
+  if ($@) {
+    $self->verbose(1,"Failed to parse META6.json file: $@");
+    File::Copy::copy $metafile, "$MLROOT/$sans.meta";
+    return;
+  }
+  close $meta_fh;
+
+  $meta->{'source-url'} = $PAUSE::Config->{PUB_MODULE_URL} . "$MLROOT/$dist";
```
ugexe: Uploading a Perl6 distribution ended up with source-url set as: http://www.cpan.org/authors/id//home/ftp/pub/PAUSE/authors/id/U/UG/UGEXE/Perl6/Text-Levenshtein-Damerau-0.1.7.tar.gz

Maybe the prefixing part of `$MLROOT` needs to be replaced with `PUB_MODULE_URL` instead of concatenating them?

PR #245, opened 2017-05-20
Fix updated source-url in Perl 6 META6.json files

ugexe:
Untested

Currently uploading a Perl6 distribution to cpan will result in a source-url such as:
http://www.cpan.org/authors/id//home/ftp/pub/PAUSE/authors/id/U/UG/UGEXE/Perl6/Text-Levenshtein-Damerau-0.1.7.tar.gz

Where the extra portion matches [$MLROOT](https://github.com/andk/pause/blob/259318c279d88b65ebfd36a8bd705a0fc00e48f6/lib/PAUSE.pm#L136) - however I am making the assumption that `$MLROOT` value is not modified.

@niner

### DIFF
--- a/lib/PAUSE/dist.pm
+++ b/lib/PAUSE/dist.pm
@@ -1085,7 +1085,7 @@ sub write_updated_meta6_json {
   }
   close $meta_fh;
 
-  $meta->{'source-url'} = $PAUSE::Config->{PUB_MODULE_URL} . "$MLROOT/$dist";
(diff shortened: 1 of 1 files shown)


niner:
Oh my, what a stupid oversight on my part. I guess this slipped through because my test script did not have $MLROOT set. Your fix looks entirely reasonable and exactly like what I actually meant there.

andk:
Thanks, applied, and deployed to live pause just now.

PR #261, opened 2018-04-19
Try using mysql_auto_reconnect which mod_perl had set under the hood

charsbar:
mysql_auto_reconnect has its own issues, but this might help for now.


andk:
merged as a7822bff98b6e840c0d766617131ba87f02d0f5f and deployed at 17:43 UTC. Thanks!

PR #263, opened 2018-04-19
Add privatelib (and cover_db) to .gitignore so that important information is not included in a commit

charsbar:


andk:
Thanks, applied.

PR #264, opened 2018-04-19
Set up signal handlers to use always Log::Dispatch::Config logger

charsbar:
so that warns and dies have timestamps.


charsbar:
The important stuff of this PR has already been merged.

PR #267, opened 2018-04-19
"Login" is a noun; the imperative verb is "log in"

karenetheridge:


karenetheridge:
Looks like this didn't make it into the mojo branch -- the templates were created before this change was applied, and wasn't merged in.  The file to edit is `pause_2017/templates/_user_menu.html.ep`.

charsbar:
@karenetheridge thanks for the nudge! I might still have a few things I need to incorporate into PAUSE on Mojo. Will look later again.

karenetheridge:
danke!

PR #273, opened 2018-04-21
misc. refactor, largely around PAUSE::dist guts

rjbs:
This was part of my general pre-cleanup for doing refactoring on the dist indexing.  Most of the heavy lifting here is reducing the way data copied from one object to another, and making it easier to get from a package up to, say, the mldistwatch object.


andk:
Merged as 0b0de75f (with reordering) and deployed. Thanks!

PR #274, opened 2018-04-21
Update README instructions for setting up private PAUSE

xdg:


xdg:
This will also need some tweaks for mojolicious.

andk:
Merged as ed836993. Thanks!

PR #275, opened 2018-04-22
Extract primeur and perms SQL to permissions manager class

xdg:
These commits pull all SQL that affects user permissions to a separate class.  Eventually, we hope to extend this to the web-side as well, but for now, it simplifies things for the mldistwatch side, including the case-conflict code that Rik is working on.


andk:
Merged and deployed at 13:49 UTC. Thanks!

PR #296, opened 2018-07-31
Link to the PAUSE operating model in 04pause

Grinnz:
Many authors see 04pause when considering uploading modules for the first time, and the PAUSE operating model contains lots of important information regarding indexing permissions.


andk:
Thanks applied and deployed yesterday.
-- 
andreas

PR #309, opened 2019-04-25
Add a file to add distname column to perms table

charsbar:
I'd like to have a new column called ```distname``` to implement permission manipulation per distribution that only holds a name part of the current ```dist``` column.


andk:
THX, merged and deployed a few minutes ago

PR #311, opened 2019-04-25
Add a script to fill packages.distname

charsbar:
Here's a script to fill packages.distname.


andk:
THX, merged an deployed and executed the script a few minutes ago

PR #313, opened 2019-04-25
misc. improvements, especially in test libraries

rjbs:
This is a raft of test improvement, plus one or two small (??) changes in behavior.  The point here is to add test tooling to make it easier to both read and write tests, especially in preparation for the rjbs/case-updating branch, which changes the behavior of case-conflicting uploads.


andk:
THX, deployed a few seconds ago

PR #315, opened 2019-04-26
Replace generated numeric ids with more human friendly ids generated from the header text

charsbar:
That replaces toc_1, toc_2 ... in ```<h3 id="...">...</h3>``` with the_pause_operating_model, version_2, 1_background and so on, to make it easier to create a link for PAUSE Operating Model document.


andk:
THX, merged and deployed a few minutes ago

PR #318, opened 2019-04-27
Use logger instead of printing out to stderr directly

charsbar:


andk:
THX, merged an deployed just a few minutes ago

PR #319, opened 2019-04-27
the binary distribution convention is long dead

### LINKED ISSUE #46: eliminate the "bindist" handling
There's code like this in `PAUSE::dist`:

```
my $binary_dist;
# ftp://ftp.funet.fi/pub/CPAN/modules/05bindist.convention.html
$binary_dist = 1 if $dist =~ /\d-bin-\d+-/i;
```

That 05 file is long gone, it seems.  I think this behavior is also long dead.

This issue is to remind me (rjbs) to kill off the code path.

rjbs:
...so remove its code.

This resolves #46


andk:

PR #320, opened 2019-04-27
when owner uploads a re-cased version of a package, update things

rjbs:
When someone with permissions to update Foo::Bar uploads foo::bar, we now update all references to Foo::Bar to refer to foo::bar.


andk:

neilb:
This makes me happy.

PR #321, opened 2019-04-27
Hide permissions per distribution pages for now

charsbar:
... until we make them fast enough under MySQL 5.1 environment


andk:

PR #322, opened 2019-04-28
Fix a too slow query under MySQL 5.1

charsbar:
This should fix that too slow query under MySQL 5.1/CentOS6.


neilb:
If you tell me when this goes live, I'm happy to test it again :-)

andk:

PR #323, opened 2019-04-28
Fix wrong parameter names and a wrong request method limitation

charsbar:
Parameters for View Permissions per distribution were stale and wrong. And also the page should allow get access.


andk:

PR #326, opened 2019-04-28
Show permissions per distribution again

charsbar:


andk:

PR #327, opened 2019-04-28
normalize all packages, not just one

rjbs:
When normalizing case, a silly mistake meant that we didn't update all packages, only one.  This fixes it, and adds a test.


andk:

Issue #229, opened 2016-07-22
users should not be able to drop firstcome without a replacement user


karenetheridge:
It looks like an author can drop firstcome permissions on a module and leave the module ownerless, even if there are other users with comaint permissions still -- for example Business::PaperlessTrans::Client.

If a user drops firstcome and doesn't appoint a successor, then ownership should pass to a placeholder user such as ADOPTME.

(I would also suggest that all existing modules in the database that have 'c' permissions but no 'f' be given an 'f' entry to ADOPTME, or perhaps a new user such as UNKNOWN.)

charsbar:
With #328, giving up primary permissions transfers their permissions to ADOPTME. If this is inconvenient, please open a new issue (and maybe refer to this). I'm closing this for now.

PR #328, opened 2019-04-28
Giving up primary permissions should mean transferring them to adoptme

### LINKED ISSUE #307: "Give up primary permissions" page is a bad idea
The "Give up primary permissions" page (https://pause.perl.org/pause/authenquery?ACTION=remove_primary) allows a user to downgrade their first-come on a module to co-maint. This should never be done, as it leaves a module in a state where there are permissions on it, but no first-come to change its permissions aside from every co-maint removing their own permissions. Instead, there should be an option to give up your first-come entirely (not assigning co-maint), and only if there are no other users with permissions. The alternative is that first-come should be transferred to someone (ADOPTME or HANDOFF if nothing else), which is already available on the "Transfer primary permissions" page.

Related: #143

charsbar:
This should close #307 and #229.


andk:

PR #329, opened 2019-04-28
indexer: cope with undef "file" in "provides" metadata

rjbs:
(by skipping those entries)


rjbs:
This MR replaces https://github.com/andk/pause/pull/71/files

andk:
THX applied and deployed a few seconds ago

PR #330, opened 2019-04-28
indexer: treat explicit provides/version as undef, but warn

rjbs:
This is an updated version of https://github.com/andk/pause/pull/68

Rejecting this for spec violation isn't trivial right now, but I support doing it in the future.  For now, this just avoids a perl warning and adds a PAUSE warning.


andk:

PR #331, opened 2019-04-28
when looking for $VERSION assignment, don't be fooled by comparisons

rjbs:
Based on, and replacing, https://github.com/andk/pause/pull/70

Now with tests.


andk:

karenetheridge:
Note: changes to  lib/PAUSE/pmfile.pm means that a new release of Parse::PMFile is necessary (and potentially a new issue raised in Module::Metadata to ensure compatible behaviour as well).

PR #332, opened 2019-04-28
Allow edit credentials when email is CENSORED

### LINKED ISSUE #139: Editing user info fails if CENSORED is in email address box
In some cases, the form is populated with "CENSORED" in the public email box.  Submitting with that value fails email address validation.  We should do one of the following:
- not populate a box with "CENSORED" and just leave it blank
- ignore "CENSORED" on submission

I looked at the code and didn't feel confident editing the email validation section myself, so this is an issue, not a PR.

charsbar:
This should fix #139


andk:

Issue #259, opened 2018-03-04
Documentation of how to upload to PAUSE from a GitHub repo tarball URL


ap:
Regarding fcee050d60cf54ca48de1df4e11f698dc70ed519 – the contents of the dead link are archived at
https://web.archive.org/web/20111202165941/http://blog.usarundbrief.com:80/?p=36

However, the URLs in it are old. Following the instructions there will still work, but instead of `.../tarball`, GitHub itself now uses `.../releases` for the tag list page and `.../archive/...` for the download link. So the example URL should now be
[https://github.com/mschilli/proc-simple-perl/__*archive*__/rel_131/Proc-Simple-1.31.tar.gz](https://github.com/mschilli/proc-simple-perl/archive/rel_131/Proc-Simple-1.31.tar.gz).

I think it makes sense to upstream this into CPAN’s own docs somewhere instead of linking Mike’s dated (if not outdated) blog post forever more.

charsbar:
Due to Dist::Zilla and its friends, people tend to keep less files in their repository than before. So just changing the URL to .../archive/... may not always work. IMHO it makes more sense to remove those sentenses, or even add some extra check to prevent inappropriate upload if it comes directly from GitHub.

karenetheridge:
We could require the presence of at least one of: META.yml, META.json,
Makefile.PL, Build.PL (which contain nearly-required information such as
author, licence, and prerequisites) in the uploaded distribution in order
for it to be indexed by PAUSE.

charsbar:
@karenetheridge META.yml and META.json are good enough when you create a distribution by Dist::Zilla etc, but are they really enough for all the current CPAN clients to configure/build/install it?

karenetheridge:
Are you saying this is too much, or not enough? We don't require a Makefile.PL today, in order for a distribution to be indexed.

charsbar:
@karenetheridge We don't require Makefile.PL/Build.PL for a distribution to be indexed, but I doubt all the current CPAN client can install a distribution that doesn't have Makefile.PL/Build.PL (some can, and it'd be nice if all would be able to do that eventually, though). So I asked you if it's really good for PAUSE to have such a distribution.

ap:
> Due to Dist::Zilla and its friends, people tend to keep less files in their repository than before. So just changing the URL to .../archive/... may not always work.

[...]

Issue #183, opened 2015-09-03
Link to upload via github URL explanation is dead


Relequestual:
When uploading a module, there is the option to do so via a URL. There is then a comment that explains it can be done directy from a github repo...

"How to use this for direct publishing from your github repository has been described by Mike Schilli in his blog"

The link however (http://blog.usarundbrief.com/?p=36) is dead. Does anyone have a copy / want to write a new explanation which could be hosted here on github or as part of the FAQ?

karenetheridge:
I don't like the idea of encouraging this practice. Part of a well-formed CPAN distribution are files that are auto-generated (e.g. by ExtUtils::MakeMaker) and are not normally included in a git repository.

Relequestual:
I DID think it was a little bit odd! Although you COULD have a dist branch for that auto-generated, although I'm not sure this would be that well used. I haven't yet investigated into command line upload options, but I'm sure that's preferable.

Maybe the solution is to update the section in question with "how to do this via cli" information?

GPHemsley:
Link is still present and still dead.

dolmen:
@GPHemsley Directly uploading to CPAN from the source repo of a Perl project is still a bad idea. You would have to commit in your repo some files that are nowadays machine generated (MANIFEST, META.yml, META.json). Also some files which are in the repo should not be in the CPAN distribution (.gitignore, .travis.yml...).

Relequestual:
@dolmen I had assumed you may have a build process and a release branch which would avoid those sorts of problems.

charsbar:
re: #259

charsbar:
The note on fetching a tarball from GitHub and the link in it are removed. Closing.

PR #333, opened 2019-04-28
Remove a note on fetching tarball from GitHub

charsbar:
as the article is already gone for years and the practice is error prone.
re #259, #183


andk:

PR #334, opened 2019-04-28
Validate email while processing request_id

### LINKED ISSUE #62: Verify email address domain name validity for ID requests
We should check email addresses in the PAUSE ID request form with Email::Valid (or equivalent), particularly including the domain name.

Invalid emails are a problem because welcome/password emails will bounce.  They may also be a sign of a bot/spam attack.

charsbar:
This should fix #62


andk:
THX, applied and deployed in the last hour

PR #335, opened 2019-04-28
use the new Module::Faker meta_munger to produce nonsense

rjbs:
CPAN::Meta sometimes protects us from making bad metadata, but in the PAUSE tests we want bad metadata sometimes!


rjbs:
Applied.

PR #336, opened 2019-04-28
Warn if a non stable release is going to be reindexed

### LINKED ISSUE #122: UX: Dev releases and reindexing menu
I just visited https://pause.perl.org/pause/authenquery?ACTION=reindex and picked from the offered file names two that matched the dev release pattern. Pause let me do this and pretended this would lead to a reindexing. But as a matter of fact nothing will happen because they are dev releases. Bad UX.

Thanks to Dave Rolsky for bringing this to my attention.

charsbar:
This should fix #122.


andk:

PR #337, opened 2019-05-03
remove remaining search.cpan.org usage in documentation

Grinnz:
No need to mention search.cpan.org anymore. Also I could not find the mentioned 'cpan-upload-http' script, so changed the references to CPAN-Uploader's cpan-upload script.


Grinnz:
I found the cpan-upload-http (https://metacpan.org/release/BRADFITZ/cpan-upload-http-2.4), because the distribution contains no modules it's not possible to find or install by name. So I think it's still better to reference CPAN-Uploader which is more used now.

andk:
Thanks, applied and deployed a few minutes ago. Thanks for help and patience,

PR #341, opened 2019-07-02
Make sure active user has a valid email

charsbar:


andk:
Thanks, merged and deployed just now!

Issue #283, opened 2018-05-23
Implement "Right to Erasure" required by GDPR


neilb:
We should provide a mechanism whereby a user can ask for their personal data to be erased.

At the moment I think we can satisfy this by blanking out the following fields of a PAUSE account:

* Full name
* ASCII name
* Public email address
* Secret email address
* Home page URL
* Password

The **full name** could be set to something like "Account Erased at user request", and perhaps a new flag added to accounts that indicates that the account has been erased.

The `00whois.xml` file would still include the account, but it would look something like this:

    <cpanid>
        <id>ZAPHOD</id>
        <fullname>Account Erased at user request</fullname>
        <introduced>123456789</introduced>
        <erased>1</erased>
    </cpanid>

This might be provided as a command that's only available to PAUSE admins, or might be available directly to end users. If the latter, it would need careful confirmation, for example like github's requiring you to enter the name of a repo, when you ask for it to be deleted.

PR #343, opened 2019-07-22
Hide the detail of deleted (or nologin) accounts

charsbar:
Eventually we need to add some measure to make sure the real data of deleted accounts in the mod table is properly deleted or anonymized, but for now, let's remove the detail of deleted (and nologin) accounts from our indices. (re: #283)


andk:
Thanks, merged and deployed ad around 01:32 UTC

PR #347, opened 2019-12-23
Fix typos

ferki:
I've noticed these minor typos and decided to send a quick PR about them.

Based on git history, the HTML file accompanying the POD seems to be autogenerated, but I'm not yet sure how exactly. I'm happy to include that step in this PR if needed, just point me in the right direction, please :)


rjbs:
Thanks!

ferki:
Thanks for merging, @rjbs!

PR #359, opened 2020-08-10
Expand operating model list of prohibitions

xdg:
Expand and standardize terminology.  Additions are:
* security
* a catch all for anything else damaging to the ecosystem

### DIFF
--- a/doc/operating-model.md
+++ b/doc/operating-model.md
@@ -149,8 +149,9 @@ Please don't abuse this.
   they are archives of everything that has ever been released to CPAN.
   Deleting a file from your author directory will not result in it
   being removed from BackPAN sites.
-* If a portion of your upload is offensive, illegal,
-  or infringes on someone else's legal rights,
+* If a portion of your upload is offensive, is illegal,
+  infringes on someone else's legal rights, poses an unwarranted security
+  risk, or is otherwise deemed damaging to the CPAN ecosystem,
   the PAUSE admins may delete a file from your author directory.
   Where appropriate,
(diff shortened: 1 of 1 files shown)


andk:
Thanks, David, highly appreciated! I've applied, version bumped, and deployed in the last few minutes.

andk:
after having slept over it I think it wasn't a good idea to accept
this PR. Blame the weather, blame my ignorance, I don't know. I wasn't
in the best shape and decided quickly instead of thoroughly.

I think we really should kill the catch all part of the pull request.
The Operating Model at its very heart had the idea that we guarantee
our users that we do not interfere with their uploads unless in
narrowly defined special cases.

Now we had an incident that showed us we had forgotten to talk about
security matters. We should react to this incident with a new version
that introduces security matters. No more, no less. When we have
another case that shows us that we missed something, we should seek
how to codify additional rules. That's what the spirit of the
Operating Model kind of promises.

I've now reverted the last two commits and would ask for a second
attempt at a rephrasing, this time strictly limited to what the
security incident taught us.

Thanks && sorry for my lapse

xdg:
With all due respect, no.

I asked for feedback 11 days ago and you had none.  I opened a PR and you accepted it without comment.  If you would like different wording, please come up with it yourself.

andk:
Absolutely understandable. Will do. Thanks for the first version anyway.

andk:
My version is in the branch https://github.com/andk/pause/tree/operating-model-add-security ; I'd deploy it in the near future if nobody raises doubts. Thanks for your thoughts and your patience

Issue #348, opened 2020-01-03
People are losing permissions (during upload?)


andk:
In November ETJ lost perms on PDL::LinearAlgebra::{Real,Trans,Special}. Six postings leading to resolution https://www.nntp.perl.org/group/perl.modules/2019/12/msg100697.html

neilb:
This also happened to `Hash::Merge`.

At the start of July, the permissions were:

    Hash::Merge,DMUEY,c
    Hash::Merge,HERMES,c
    Hash::Merge,MNEYLON,c
    Hash::Merge,REHSACK,f

Then HERMES did two releases, 0.301 and 0.302, and one of those releases seems to have resulted in all permissions being deleted. On the evening of 2020-07-30 (UTC), I checked and confirmed that there were no permissions on `Hash::Merge` at all.

I've added them back.

oalders:
I’m wondering if this happened to me with WWW-YouTube-Download. I tried to upload a new release last month and my perms were gone:

>  User: OALDERS (Olaf Alders)
>  Distribution file: WWW-YouTube-Download-0.63.tar.gz
>  Number of files: 44
>  *.pm files: 1
>  README: WWW-YouTube-Download-0.63/README.md
>  META-File: WWW-YouTube-Download-0.63/META.json
>  META-Parser: Parse::CPAN::Meta 1.4414
>  META-driven index: no
>  Timestamp of file: Tue Jun 23 21:14:59 2020 UTC
>  Time of this run: Tue Jun 23 21:36:41 2020 UTC
> 
> ERROR: Database error occurred during index update
> 
> 
> This distribution name will only be indexed when uploaded by users with
> permission for the package WWW::YouTube::Download. Either someone else
> has ownership over that package name, or this is a brand new
> distribution and that package name was neither listed in the 'provides'
> field in the META file nor found inside the distribution's modules.
> Therefore, no modules will be indexed.
> 
> Further details on the indexing attempt follow.
> 
> 
> The following packages (grouped by status) have been found in the distro:
> 
> Status: Permission missing
> ==========================
> 
>     module : WWW::YouTube::Download
>     version: 0.63
>     in file: WWW-YouTube-Download-0.63/lib/WWW/YouTube/Download.pm
>     status : Not indexed because permission missing. Current registered
>             primary maintainer is XAICRON. Hint: you can always find
>             the legitimate maintainer(s) on PAUSE under "View
>             Permissions".

neilb:
[...]

PR #364, opened 2020-10-18
Tests and proposed fix for #348, where first-come can be lost

neilb:
This is my proposed fix for #348:

If a package has one author with first-come on a package
(but only with an entry in primeur, not in perms),
and someone else has co-maint (entry in perms).
If the co-maint does the next release, then the entry in primeur
is getting deleted, so it ends up with no first-come.

I added test `t/losing-first-come.t` which failed until canonicalize_module_casing() in `PAUSE::PermsManager`
was fixed.
I added another test `t/propagating-permissions.t`, which explicitly tests the expected behaviour
on propagating the permissions on the lead package to other packages, regardless of who releases.
This was so there was additional testing on the fix I made.

Notes: the mldistwatch-* and other tests all pass, though the ones that require Test::mysqld were skipped, as I don't have MySQL installed. Shout if you want me to do that.

I think @andk, @charsbar, and @rjbs should review the proposed change, as I'm not familiar enough with all the assumptions that might be made elsewhere, with respect to `primeur` and `perms`.

### DIFF
--- a/lib/PAUSE/PermsManager.pm
+++ b/lib/PAUSE/PermsManager.pm
@@ -200,16 +200,22 @@ sub canonicalize_module_casing {
   my $dbh = $self->dbh_callback->();
   my $users = $dbh->selectall_arrayref(
     qq{
-      SELECT perms.userid, COUNT(primeur.userid) AS is_primary
-      FROM perms
-      LEFT JOIN primeur
-        ON  LOWER(primeur.package) = LOWER(perms.package)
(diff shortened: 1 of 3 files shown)


andk:
Thank you, Neil, this looks awesome. I merged it to see it running through travis. I'll deploy it when I have slept over it a few times and nobody vetos.

andk:
Deployed now on production. Thanks again, closing issue.

PR #378, opened 2022-01-16
mldistwatch: point to canonical URL for 02packages

rjbs:
...not the perl.com redirect, which we don't control.

### DIFF
--- a/lib/PAUSE/mldistwatch.pm
+++ b/lib/PAUSE/mldistwatch.pm
@@ -692,7 +692,7 @@ sub rewrite02 {
       unless $numlines >= $PAUSE::Config->{ML_MIN_INDEX_LINES};
 
     my $header = qq{File:         02packages.details.txt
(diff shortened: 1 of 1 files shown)


abh:
I imagine it can't be changed to a https:// URL, but maybe if we're changing this we could add a canonical https URL as well that clients with support can use?

andk:
>>>>> On Sun, 16 Jan 2022 18:16:13 -0800, Ask Bjørn Hansen ***@***.***> said:

  > I imagine it can't be changed to a https:// URL, but maybe if we're
  > changing this we could add a canonical https URL as well that clients
  > with support can use?

FWIW, from my internet connection this address already works:

    https://www.cpan.org/modules/02packages.details.txt

I would not mind putting this in place.

For now I have applied and rolled out Ricardo's pull request.

Thanks,
-- 
andreas

rjbs:
I didn't even consider the http/https question, and assume it's fine, but am not sure what the implications would be.  What, if anything, uses that field?  I could not say!

Issue #352, opened 2020-06-24
Add SHA-2 Support


dweekly:
As of 2020 SHA-1, sadly, no longer provides strong signing assurance. SHA1 was [cracked in 2017](https://shattered.it/). As of 2020 [it costs under $50k to crack](https://eprint.iacr.org/2020/014.pdf), [causing OpenSSH to deprecate SHA-1](http://www.openssh.com/txt/release-8.3).

We could use mature, stronger hashing algorithms like SHA-2 (SHA-256), or SHA-3. A variety of Perl libraries are available already that implement SHA-2 and SHA-3. An ideal would be for us to assume the need to change signature standards periodically (on the order of about once a decade). Of existing Perl SHA-2 and SHA-3 implementations, Digest::SHA already includes SHA-256 support and is already part of perl core as of 5.9.3, so seems the safest bet.

PAUSE need to be updated to emit these new signatures, as well as (lower-priority) also encouraging individual authors who have included signatures in their modules to update the quality of those signatures.

The Module::Signature request to support SHA256 hashes is in [issue 28](https://github.com/audreyt/module-signature/issues/28).

This request is part of a larger effort to [Secure Perl](https://docs.google.com/document/d/1DRkiCJhJu4RDI0u_JppBpFa0djouskxEyNHax912U_w/edit#).

dweekly:
Current plan is to base implementation on https://github.com/andk/pause/pull/211/files

dweekly:
Rad - thanks!

PR #379, opened 2022-02-01
Use sha256 instaed of sha1 or md5sums

mbeijen:
This closes https://github.com/andk/pause/issues/352

I decided to remove the sha1sum and the md5sum from the mails; if we think md5sums and sha1sums are no longer 'safe' then we should not provide it to users; otherwise they might still be using those to compare their files! I made it separate commits so you can do otherwise as you see fit; but feel free to squash them or if you want of course I can also do that.

BTW I came here via this comment on the release of a new Perl: https://www.nntp.perl.org/group/perl.perl5.porters/2022/01/msg262480.html

I found out that the release manager pastes the shasum from the pause email; so I went looking how to provide sha256 there, and that lead to this PR. 

Thanks in advance!

### DIFF
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -52,6 +52,7 @@ WriteMakefile(
     LWP::UserAgent
     LWP::Protocol::https
     Digest::MD5
+    Digest::SHA
     Digest::SHA1
     Mail::Mailer
     Mail::Send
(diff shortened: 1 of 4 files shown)


andk:
Thanks, LGTM, merged and deployed just a few minutes ago.

mbeijen:
Awesome, thanks

PR #386, opened 2023-01-15
LWP::UserAgent: require TLSv12, stop trying to use SSLv3

rjbs:
I'm not sure this is right, but if you look at today's logs, you'll see that PAUSE got stuck unable to fetch an upload because PAUSE wanted to use old SSLv3.

```
2023-01-14 00:21:17 $3234 v1049: Info: Requesting a GET on uri [https://www.perturb.org/tmp/Date-Parse-Modern-0.1.tar.gz] (paused:671)
2023-01-14 00:21:18 $3234 v1049: Alert: nosuccesscount[6] error[Can't connect to www.perturb.org:443 (certificate verify failed)] (paused:701)
```

This is probably right-er, but may be too much?

### DIFF
--- a/bin/paused
+++ b/bin/paused
@@ -155,7 +155,9 @@ sub loop () { # we're NOT called as a method
 
   our $UA;
   unless ($UA) {
-    $UA =  LWP::UserAgent->new;
+    $UA =  LWP::UserAgent->new(
+      ssl_opts => { SSL_version => "TLSv12" },
+    );
     $UA->timeout($PAUSE::Config->{TIMEOUT}) if $PAUSE::Config->{TIMEOUT};
   }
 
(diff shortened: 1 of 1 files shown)


haarg:
This seems like a good change.

We should also consider changing the defaults in LWP::Protocol::https, or maybe even Net::SSLeay.

rspier:
"Recent" IO::Socket::SSL has a better default: https://github.com/noxxi/p5-io-socket-ssl/blob/master/lib/IO/Socket/SSL.pm#L202

```perl
    SSL_version => 'SSLv23:!SSLv3:!SSLv2', # consider both SSL3.0 and SSL2.0 as broken
```
since https://github.com/noxxi/p5-io-socket-ssl/commit/42fd97a8a445060b621f98eb7edecca77fe38796 (8 years ago)

Although even that may not be sufficient as some hosts aren't supporting TLS1.0 either.

haarg:
I've filed noxxi/p5-io-socket-ssl#122 regarding changing IO::Socket::SSL to only support TLS 1.2 and newer.

rjbs:
I think the right fix is to update our required IO::Socket::SSL instead.

rjbs:
PAUSE's SSL has been updated, so this should be resolved.  Let us know if not!

PR #387, opened 2023-04-22
Docker compose

charsbar:
This PR adds docker-compose files to help develop PAUSE. It is not to suggest changing how to deploy PAUSE or anything else. `docker-compose up` runs paused, plack app behind nginx, mailhog to see emails that PAUSE sends. It doesn't run cron scripts (at least for now). No FTP nor Rsync servers. TESTADMIN and TESTUSER accounts (the passwords are both "test") are predefined.


oalders:
@charsbar I probably help with `Mojolicious::Plugin::WithCSRFProtection`, if you need something fixed.

PR #389, opened 2023-04-27
use Email::Sender rather than Mail::Send

rjbs:
We already use Email::Sender in a number of places, and this converts the rest.  This is being done so that the mail transport can be configured a single way for testing, and later maybe in configuration for various Docker environments.


andk:
Thanks, applied and deployed

PR #392, opened 2023-04-28
mldistwatch: forbid tarbombs from being indexed

rjbs:
Previously, you could upload a tarball with "Foo.pm" in the root and we would index that.  From now on, dists must have a top-level directory, and only that one directory.


haarg:
Does this handle a tarball with a top level '.' directory properly?

rjbs:
No!

andk:
@rjbs, could you review and resolve conflicts?

PR #393, opened 2023-04-28
refactor "may file be overwritten" code (and add Markdown)

rjbs:
This came up during a discussion of "what files can be overwritten?"  I looked it up, found it in a place that is harder to test, and moved it to somewhere simpler.

I also…
1. removed an ancient exception that is long-unused
2. added Markdown extensions
3. permit README.gz, mostly as an accident of refactoring

### DIFF
--- a/lib/PAUSE.pm
+++ b/lib/PAUSE.pm
@@ -553,6 +553,12 @@ sub isa_regular_perl {
   return scalar $filename =~ $ISA_REGULAR_PERL;
 }
 
+sub may_overwrite_file {
+  my ($filename) = @_;
+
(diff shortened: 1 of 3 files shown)


lib/PAUSE.pm
```diff
@@ -553,6 +553,30 @@ sub isa_regular_perl {
   return scalar $filename =~ $ISA_REGULAR_PERL;
 }
 
+sub may_overwrite_file {
+  my ($filename) = @_;
+
+  my $is_archive = $filename =~ s/\.(gz|bz2)$//i;
```
ilmari, line 559: This variable seems unused?

PR #397, opened 2023-04-28
Fix DOM structure of 04about

### LINKED ISSUE #396: pause_04about is broken
https://pause.perl.org/pause/query?ACTION=pause_04about is broken now because of https://github.com/andk/pause/pull/385 . Needs to fix its DOM structure.

charsbar:
This fixes #396


andk:
Thanks, applied and deployed

Issue #192, opened 2015-12-20
very small numeric versions can break indexing


rjbs:
Today, EXODIST uploaded a new dist with: `$VERSION = 0.000001`;

Sure, this *should* have been a string, but it should have worked anyway, right?  Well, here are logs you get if you index his upload in a local clone of PAUSE:

```
Untarred '/var/folders/tp/xbk5yqfj7vv86jjcgk_cp4wh0000gq/T/UlbG1vMSWa/cpan/authors/id/L/LO/LOCAL/Test2-Essentials-0.000001.tar.gz'
Found 10 files in dist L/LO/LOCAL/Test2-Essentials-0.000001.tar.gz, first Test2-Essentials-0.000001/Changes
Finished with pmfile[Test2-Essentials-0.000001/lib/Test2/Essentials.pm]
Result of normalize_version: sdv[1e-06]
error with version in Test2-Essentials-0.000001/lib/Test2/Essentials.pm: Invalid version format (non-numeric data) at /Users/rjbs/code/hub/pause/lib/PAUSE/pmfile.pm line 537.
st[$VAR1 = undef;]
Sent "indexer report" mail about LOCAL/Test2-Essentials-0.000001.tar.gz
 Uploading user has no permissions on package Test2::Essentials
```

PAUSE::pmfile::normalize_version converts the number 0.000001 to `1e-06` by using Dumpvalue.

Later, the code will do this:  `version->new("1e-06")->numify`

version.pm can't handle versions in scientific notation, and dies.

There are a lot of things here being done in a way that I'd call wrong, but I'm hesitant to fix most of them without a lot of careful checking.  I think the simple fix might be, after this line:

```
    my $sdv = $dv->stringify($v,1); # second argument prevents ticks
```

Add something like this pseudocode:

```
    if ($sdv =~ /scientific-notation/) {  $sdv = sprintf "%.30f", $sdv; $sdv =~ s/0+\z//; }
```

Awful, but better.

(On a related note, we do not detect or report versions that are too wide to fit in the index database slot for version.)

wolfsage:
Something like...
```
if ($sdv =~ /e-(\d+)/) {
  my $precision = $1;
  $sdv = sprintf "%.*f", $precision, $sdv;
  print "S: $sdv\n";
}
```

PR #399, opened 2023-04-28
on a very small (or large) version, do not index

rjbs:
Really, this should become a useful user message, but it's painfully difficult to get that error up through the call stack.  This code needs (in my opinion) to be rearchitected to make this sort of thing simple.

In the meantime, the code now detects this case specifically, so we can do something about it later.

Meant to help with #192, but won't help enough until we can get the message passed upward.


andk:
@rjbs, could you please review and fix the conflicts?

PR #400, opened 2023-04-28
Test::PAUSE::Web: Use ->decoded_content

wolfsage:
->content might be compressed! (It is on newer Mojo apparently!)


andk:
Thanks, applied

PR #401, opened 2023-04-28
refuse to index dists with no META.yml or META.json

### LINKED ISSUE #164: require META file for new uploads to be indexed
As discussed at QAH, April 18th.

rjbs:
…as we planned to do years ago.  Fixes #164


andk:

PR #405, opened 2023-04-29
indexing context: make it easier to log and report on indexing status

rjbs:
WIP


lib/PAUSE/Indexer/Errors.pm
```diff
@@ -0,0 +1,121 @@
...
+};
+
+public_error no_distname_permission => {
+  header  => 'missing permissions on distname package',
+  body    => sub {
+    my ($dist) = @_;
+
+    my $pkg = $dist->_package_governing_permission;
+
+    return <<~"EOF"
```
wolfsage, line 48: NO_DISTNAME_PERMISSION used to talk about:
```
   push @m, $tf->format(qq[This distribution name will only be indexed
      when uploaded by users with permission for the package $pkg.
      Either someone else has ownership over that package name, or
      this is a brand new distribution and that package name was neither
      listed in the 'provides' field in the META file nor found
      inside the distribution's modules.  Therefore, no modules
      will be indexed.]);
```

And when...
 `        # Perhaps they forgot a pm file matching the dist name`
...then it would say the paragraph you have below.

Why is this changing here to no longer have the former case? What happens in the new code if I upload a dist someone else already owns but I do have a matching .pm name for the dist?

lib/PAUSE/dist.pm
```diff
@@ -410,214 +468,71 @@ sub mail_summary {
...
-          different distribution name could be chosen to reflect an
-          actually included package name (eg. $inxpkg_eg).\n});
-    }
-
-      for my $p ( keys %$inxst ) {
-          next unless
-            $inxst->{$p}{status} == PAUSE::mldistwatch::Constants::OK;
-          $inxst->{$p}{status} = PAUSE::mldistwatch::Constants::EDISTNAMEPERM;
-          $inxst->{$p}{verb_status} =
-            "Not indexed; $author not authorized for this distribution name";
```
wolfsage, line 453: It seems like a user won't see a not authorized error anymore?

lib/PAUSE/mldistwatch.pm
```diff
@@ -347,31 +349,31 @@ sub _do_the_database_work {
...
 
     my $main_pkg = $dio->_package_governing_permission;
 
     if ($self->permissions->userid_has_permissions_on_package($dio->{USERID}, $main_pkg)) {
       $dbh->commit;
     } else {
-      $dio->alert("Uploading user has no permissions on package $main_pkg");
-      $dio->{NO_DISTNAME_PERMISSION} = 1;
+      $ctx->alert("Uploading user has no permissions on package $main_pkg");
+      $ctx->add_dist_error(ERROR('no_distname_permission'));
```
wolfsage, line 376: Maybe all add_dist_error calls should actually be aborts?

wolfsage (changes requested):
Wild!

I think we should probably test this on the same version of Perl pause is running on to be sure we haven't introduced more incompatibilities (like indented heredocs).

I mostly understand it but it's a lot to reason about.

I think this commit message needs rewording:
```
Code that wants to abort can $ctx->abort_indexing, which adds a
dist_error uses an exception to stop processing the dist further. 
```
"which adds a dist_error uses an exception" I can't seem to parse

And the final commit regarding attempting to avoid partial indexing doesn't quite make sense. I'm reasonably sure that already happens inside of a transaction.

do_the_database work starts a transaction then does:
 * examine_pms->index_by_files->examine_fio->examine_pkg->give_regdowner_perms
Or
 * examine_pms->index_by_meta->examine_pkg->give_regdowner_perms
while in the transaction.

wolfsage:
A bunch of code can also be removed if you wish (https://github.com/rjbs/git-branchmanager)

lib/PAUSE/Indexer/Errors.pm
```diff
@@ -0,0 +1,121 @@
...
+  $ERROR{$name} = {
+    ident   => $name,
+    public  => 1,
+    %$arg,
+  };
+}
+
+public_error blib => {
+  header  => 'archive contains a "blib" directory',
+  body    => <<~'EOF'
```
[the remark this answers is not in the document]
rjbs: 😢

It's v5.16, but I've removed indented heredocs (and 😢 postfix deref).

lib/PAUSE/Indexer/Context.pm
```diff
@@ -0,0 +1,99 @@
...
+  my ($self) = @_;
+
+  return map {; @$_ } values $self->_package_warnings->%*;
+}
+
+sub warnings_for_package {
+  my ($self, $package_name) = @_;
+
+  return grep {; $_->{package} eq $package_name }
+         map  {; @$_ } values $self->_package_warnings->%*;
```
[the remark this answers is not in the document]
rjbs: Done!

lib/PAUSE/Indexer/Errors.pm
rjbs: Restored!

lib/PAUSE/dist.pm
rjbs: They'll see it in the whole-dist context, and there is no longer a per-module error in the "no distname permissions" case.

lib/PAUSE/mldistwatch.pm
rjbs: These two can't (yet) because they happen outside the error handler.

rjbs:
I reworded the awkward commit.  I also removed INDEX_WARNING stuff, which led me to find that the prelude-to-warnings text was not being included.

rjbs:
@wolfsage As discussed today, I think this is reviewable.

lib/PAUSE/Indexer/Context.pm
```diff
@@ -0,0 +1,172 @@
...
+  push @{ $self->_dist_errors }, $error;
+
+  return $error;
+}
+
+sub dist_errors {
+  my ($self) = @_;
+  return @{ $self->_dist_errors };
+}
+
```
wolfsage, line 159: nit:

`sub alert` to add an alert.
`sub all_alerts` to get alerts

`sub add_dist_error` to add a dist error
`sub dist_errors` to get dist errors

These two things are very similar but have slightly different worded interfaces

lib/PAUSE/Indexer/Context.pm
```diff
@@ -0,0 +1,172 @@
...
+has dist_errors => (
+  is      => 'bare',
+  reader  => '_dist_errors',
+  default => sub {  []  },
+);
+
+sub add_dist_error {
+  my ($self, $error) = @_;
+
+  $error = ref $error ? $error : { ident => $error, message => $error };
```
wolfsage, line 147: Later code does:
```
  my @dist_errors = $ctx->dist_errors;

  for my $error (@dist_errors) {
    my $header = $error->{header};
    my $body   = $error->{body};
```

But if you say `add_dist_error("foobar")`, there will be no header or body. We should probably just drop the string support and always expect the object/hashref

lib/PAUSE/Indexer/Context.pm
```diff
@@ -0,0 +1,172 @@
...
+  my ($self) = @_;
+
+  my %stash = %{ $self->_package_status };
+  return @stash{ sort keys %stash };
+}
+
+sub abort_indexing_package {
+  my ($self, $package_obj, $error) = @_;
+
+  $Logger->log("abort indexing $package_obj->{PACKAGE}");
```
wolfsage, line 96: Does the `$error` get logged anywhere?

lib/PAUSE/Indexer/Context.pm
```diff
@@ -0,0 +1,172 @@
...
+  return $error;
+}
+
+sub dist_errors {
+  my ($self) = @_;
+  return @{ $self->_dist_errors };
+}
+
+sub abort_indexing_dist {
+  my ($self, $error) = @_;
```
wolfsage, line 161: Does the `$error` get logged anywhere?

lib/PAUSE/package.pm
```diff
@@ -257,21 +241,13 @@ sub examine_pkg {
 
   # should they be cought earlier? Maybe.
   # but as an ultimate sanity check suggested by Richard Soderberg
-  if ($self->_pkg_name_insane) {
-      $Logger->log("package[$package] name seems illegal");
-      delete $self->{FIO};    # circular reference
```
wolfsage, line 262: We lost a lot of circular reference cleanup in this PR. Is this going to cause leaks?

lib/PAUSE/package.pm
```diff
@@ -293,81 +270,45 @@ sub examine_pkg {
...
-
-        Note: the indexer is running in a Safe compartement and cannot
-        provide the full functionality of perl in the VERSION line. It
-        is trying hard, but sometime it fails. As a workaround, please
-        consider writing a META.yml that contains a 'provides'
-        attribute or contact the CPAN admins to investigate (yet
-        another) workaround against "Safe" limitations.)},
-
-                              );
+          # TODO: get $err->{openerr} back in here, I guess?
```
wolfsage, line 273: Is this todo something to worry about?

lib/PAUSE/package.pm
```diff
@@ -293,81 +270,45 @@ sub examine_pkg {
...
+          $ctx->abort_indexing_package($self, PKGERROR('version_openerr'));
       }
-      delete $self->{FIO};    # circular reference
-      return;
+
+      # TODO: get $err->{line} back in here, I guess?
```
wolfsage, line 277: is this todo something to worry about?

lib/PAUSE/package.pm
```diff
@@ -603,143 +489,131 @@ has the same version number and the distro has a more recent modification time.}
...
-               qq{The PAUSE indexer could not store the indexing
-result in the DB due the following error: C< $dbherrstr >.
-Please report the case to the PAUSE admins at modules\@perl.org.},
-              );
-      }
+  my $rows_affected = eval {
+      $dbh->do(
+          q{
+            UPDATE  packages
+            SET     package = ?, version = ?, dist = ?, file = ?,
```
wolfsage, line 549: I think you lost an `lc_package = ?` ... `$package, lc $package, $pp->{version}` in a rebase

lib/PAUSE/package.pm
```diff
@@ -856,17 +729,18 @@ sub checkin {
...
       # we hear for the first time about this package
-      $self->insert_into_package;
+      $self->insert_into_package($ctx);
   }
 
-  my $status = $self->get_index_status_status;
-  if (! $status or $status == PAUSE::mldistwatch::Constants::OK) {
-      $self->checkin_into_primeur; # called in void context!
-  }
+  # my $status = $self->get_index_status_status($ctx);
```
wolfsage, line 738: Why are tehse conditionals commented out? Because we won't reach this if the contexts abort? Maybe remove them or add a comment?

wolfsage (commented):
*checks clock* 2h6m. *whew*.

There's.... a lot here. I've only looked at the code, not played with it. I've left a number of questions.

I think there's at least one legitimate bug which implies more tests could be written, and some code that was either lost in a rebase, or didn't make sense to keep in (lc_package in an UPDATE where it's probably not needed).

I question whether we've introduced mem leaks by not deleting `$SELF->{DIO}` or FIO everywhere like we used to.

I'd like to play with this and test it on a running inabox, but I cannot do that tonight. Wednesay perhaps if you want to wait more.

Otherwise, seems like good improvements, for the bits I could understand the changes without my brain melting

lib/PAUSE/Indexer/Errors.pm
```diff
@@ -0,0 +1,121 @@
...
+    another indexing attempt be made by logging into https://pause.perl.org/
+    EOF
+};
+
+sub ERROR {
+  my ($ident) = @_;
+
+  my $error = PAUSE::Indexer::Errors->error_named($ident);
+  unless ($error) {
+    Carp::confess("requested unknown error: $ident");
```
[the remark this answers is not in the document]
rjbs: Loaded, thanks!

lib/PAUSE/package.pm
rjbs: I don't _think_ so, but it almost doesn't matter:  mldistwatch does not make a lot of these, and it doesn't run persistently.  Still I'll give it a think.

lib/PAUSE/package.pm
```diff
@@ -707,6 +677,66 @@ Please report the case to the PAUSE admins at modules\@perl.org.},
 
 }
 
+sub __do_regular_perl_update {
+    my ($self, $ctx, $old_row, $arg) = @_;
+
```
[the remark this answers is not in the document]
rjbs: Ugh, this commit is a mess, huh?

lib/PAUSE/Indexer/Context.pm
rjbs: Fixed.  (I made `->alert` be `->add_alert`)

lib/PAUSE/Indexer/Context.pm
rjbs: I think you're confused, but maybe I'm confused.

The ternary expression says "if $error was a reference, leave it one.  otherwise, turn it into one where it's both ident and message are the string that was $error".  So, whether you pass a string or reference to `add_dist_error`, you get a reference stored in `all_dist_errors`.  The "later code" you cite will get those references and treat them as references.

lib/PAUSE/Indexer/Context.pm
rjbs: Ha, no.  I was going to say, "Yes, the next line calls `_set_package_error` which logs it."  But it only logs that there _is_ an error status.  Fixed in new commit.

lib/PAUSE/package.pm
rjbs: I believe that `lc_package` can't change here.  Passing it in the update seems pointless.  We're finding the row for `lc_package = 'dist::zilla'` and we might update its `package` field to `DIST::ZiLLA` or `Dist::Zilla` or `Dist::zillA`, but no matter what `lc_package` has to remain constant.

Agreed?

lib/PAUSE/package.pm
rjbs: Thanks, I've removed them.

lib/PAUSE/Indexer/Context.pm
rjbs: Yes, the next line calls `add_dist_error` which logs $error.

lib/PAUSE/package.pm
```diff
@@ -707,6 +677,66 @@ Please report the case to the PAUSE admins at modules\@perl.org.},
...
+    my ($opack, $oldversion, $odist, $ofilemtime, $ofile) = @$old_row{
+      qw( package version dist filemtime file )
+    };
+
+    my $older_isa_regular_perl = $arg->{older_isa_regular_perl};
```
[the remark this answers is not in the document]
rjbs: Okay, I have been putting off serious work on this MR because this was clearly the biggest thing to do: test indexing perls.  I will now attempt a basic set of tests.  Wish me luck.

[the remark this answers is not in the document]
rjbs: I added new tests (and test helpers) in #510 to test the uncovered code — this and a few other pieces, thanks to [coveralls](https://coveralls.io/jobs/141980705/source_files/25495601820).  Then I rebased this branch on that one and fixed the tests that had been broken.

lib/PAUSE/package.pm
wolfsage: Yup

lib/PAUSE/Indexer/Context.pm
wolfsage: This doesn't have to do with whether or not later code expects references, it's what that later code expects the references to **contain**.

As I said, later code wants a `header` and a `body` key, and doesn't care at all about `ident` or `message`. Here's the full snippet of what later code wants: 

```
  my @dist_errors = $ctx->all_dist_errors;

  for my $error (@dist_errors) {
    my $header = $error->{header};
    my $body   = $error->{body};
    $body = $body->($self) if ref $body;

    push @m, "## $header\n\n";
    push @m, $tf->format($body), qq{\n\n};

    $status_over_all = "Failed";
  }
```

If add_dist_error is called with a string, and we ever somehow do a mail summary send, it seems to me we'll push:
```
## \n\n

```
(for header and body respectively, as they'll be undef)
and generate two warnings.

wolfsage: Relatedly, if we ever do send a public summary, it seems like we don't attempt to filter out non-public errors! We shoudl?

lib/PAUSE/Indexer/Errors.pm
```diff
@@ -124,30 +141,48 @@ EOF
...
-  body    => <<'EOF',
-Not indexed because package $opack in file $ofile has a dual life in $odist.
-The other version is at $oldversion, so not indexing seems okay.
+  body    => sub {
+    my ($old) = @_;
+
+    _assert_args_present(db_conflict => $old, [ qw(package file dist version) ]);
+
+    return <<"EOF";
+Not indexed because package $old->{pack} in file $old->{file} has a dual life
```
wolfsage, line 153: You expect 'package' but you're using 'pack' so this throws a warning and doesn't fill in

lib/PAUSE/Indexer/Errors.pm
```diff
@@ -124,30 +141,48 @@ EOF
...
-$odist. Although the other package is at version [$oldversion], the indexer
-lets the other dist continue to be the reference version, shadowing the one in
-the core.  Maybe harmless, maybe needs resolving.
+  body    => sub {
+    my ($old) = @_;
+
+    _assert_args_present(db_conflict => $old, [ qw(package file dist version) ]);
+
+    return <<"EOF";
+Not indexed because package $old->{pack} in file $old->{file} seems to have a
```
wolfsage, line 168: You expect 'package' but you're using 'pack' so this throws a warning and doesn't fill in

wolfsage (commented):
a few small things to fix and our previous conversation to mull over, but I think this is basically there

lib/PAUSE/Indexer/Context.pm
rjbs: Okay, I've fixed this.

lib/PAUSE/package.pm
rjbs: I think it's okay to leave like this, but … I dug into this whole hunk of code and HOO BOY.  This is a whole 'nother project for a whole 'nother time.

rjbs:
I've added commits from [PAUSE::Indexer::Errors: fix keys from $old](https://github.com/andk/pause/pull/405/commits/7996eec63e46b8a576135b1bbe4025e19648c96b) to [PAUSE::dist: do not include private errors in report](https://github.com/andk/pause/pull/405/commits/6758a601824d0acb9edde978ac58e36f64187c3a) which I think address the last(???) outstanding thing.  I'm going to look into coverage a bit more.

rjbs:
I added code to let me get an mbox of all mail sent during testing, which I then inspected in my MUA. Looks pretty good.

rjbs:
I added just a bit more test coverage for a notable but never-called-during-tests method.

Issue #225, opened 2016-07-09
wishlist: indicate which files have modules in the index


karenetheridge:
For the "show my files" and "delete files" views, it would be helpful to add an icon next to files which have representation in the 02packages index, as an indicator of "if you delete this file, the index will be affected".  In other words, things without that icon are safe to delete.

Issue #350, opened 2020-04-07
Warn users on attempting to delete all versions of a distribution


Grinnz:
Deleting all versions of a distribution renders it uninstallable through normal processes, only accessible from backpan. A warning should be displayed when this is requested. Related to #204

rjbs:
From PTS 2023:  We think that warning about deleting *all* versions is complex and not all that useful.  (More below.)  On the other hand, a good warning (we think) is:  "You are deleting one or more files that appear in 02packages."

andk:
Thanks, resolved in a different way that seems to meet the objective quite nicely (see #412). Already deployed on production pause.

PR #412, opened 2023-04-30
Show [indexed] if the file is in the 02packages index

charsbar:
This PR fixes https://github.com/andk/pause/issues/225


charsbar:
Added a fix for https://github.com/andk/pause/issues/350 as well.

PR #414, opened 2023-04-30
remove "make_user_account.pl": long dead

rjbs:
...and yet dangerous!  We do not want to ever accidentally run adduser.


rjbs:

Issue #417, opened 2023-04-30
Confused entifying when asking an admin to check an account request


neilb:
When someone applies for a new PAUSE account, an admin (currently always me) goes to the "Manage a registration request (alpha)" link.

If you hit the "Insert with soundex care" button, you may get a table with existing accounts that might be the same person. E.g.

![new-user-entify-issue](https://user-images.githubusercontent.com/478461/235348311-4421bb28-b4e3-4653-845d-d0c93037c9a2.png)

As you can see, there is some weird entifying going on. `&#160;` is `&nbsp;`. It's only me doing this, and it's not bothered me enough to go fix it, but given you're all hacking away right now, if someone could fix this, that would be nice / appreciated :-)

PR #418, opened 2023-04-30
Do not double escape &#160;(nbsp)s

charsbar:
This should fix https://github.com/andk/pause/issues/417


neilb:
Thankyou Ishigaki-san – I just used this page, and benefited from this change.

PR #424, opened 2023-04-30
DO NOT MERGE! No more deadlocks

### LINKED ISSUE #406: Pause Deadlocks/database timeouts
I was able to reproduce the pause deadlocks and maybe the db timeouts on a pause docker image.

I believe the core problem is this:

Inside of a transaction (that's maaaaaaaaaaaybe a bit too long running) we do:
```
INSERT INTO primeur ...
```
And another process does this (outside of a transaction):
```
DELETE FROM perms WHERE LOWER(package) = LOWER(?)
```

The LOWER(package) prevents mysql from using an index lookup, which means it scans the *entire table/index* and locks way more than it needs to, and gets stuck behind the transaction with the INSERT.

I think to solve this we need to either...
1. Get rid of all duplicate rows due to historical bad data, then modify the column to have default mysql collation which will mean 'X' = 'x' in searches so we can do plain index lookups
2. Add another column "package_lower" which is the same as package but always lowercase, and use that
3. Do a select before any update and delete to retrieve the correct case package names and then do the updates/deletes with that data

Option 1 is probably best, 2 is okay, 3 sucks

wolfsage:
Fixes #406 

This cannot be done until the database is cleaned up of duplicates however. (Soon, I hope!)

To use this, we'll first have to do:
```
ALTER TABLE perms MODIFY `package` char(245) NOT NULL DEFAULT '';
ALTER TABLE packages MODIFY `package` varchar(128) NOT NULL DEFAULT '';
ALTER TABLE primeur MODIFY `package` char(245) NOT NULL DEFAULT ''; 
```

### DIFF
--- a/doc/mod.schema.txt
+++ b/doc/mod.schema.txt
@@ -218,7 +218,7 @@ CREATE TABLE p6provides (
 /*!40101 SET @saved_cs_client     = @@character_set_client */;
 /*!40101 SET character_set_client = utf8 */;
 CREATE TABLE packages (
-  package varchar(128) CHARACTER SET latin1 COLLATE latin1_bin NOT NULL DEFAULT '',
+  package varchar(128) NOT NULL DEFAULT '',
   version varchar(16) NOT NULL DEFAULT '',
   dist varchar(128) NOT NULL DEFAULT '',
   distname varchar(128) NOT NULL DEFAULT '',
@@ -240,7 +240,7 @@ CREATE TABLE packages (
 /*!40101 SET character_set_client = utf8 */;
 CREATE TABLE perms (
   c int(10) unsigned NOT NULL AUTO_INCREMENT,
-  package char(245) CHARACTER SET latin1 COLLATE latin1_bin NOT NULL DEFAULT '',
+  package char(245) NOT NULL DEFAULT '',
   userid char(10) NOT NULL DEFAULT '',
   PRIMARY KEY (c),
   UNIQUE KEY pack_user (package,userid),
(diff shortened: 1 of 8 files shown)


andk:
Hi Matthew,
could you please consider the scenario that people change case within their releases over time? Will the code in the MR properly deal with this? I believe proper handling of this would be a rejection. From reading the idea to make the package column case-insensitive I get the impression that mysql might lose control of the correct case sensitive package names. The idea came up while talking with @neilb about RLAUGHLIN/UDPmsg-0.10.tar.gz vs RLAUGHLIN/UDPmsg-0.11.tar.gz and about SHERZODR/CGI-Session-3.95 vs MARKSTOS/CGI-Session-4.48. There are more such examples in the history of the CPAN, where individual files within a package changed case. Breaks code on case-sensitive file systems. It follows quite naturally from working on a case insensitive filesystem, that developers make such changes, but it is something a repository like CPAN should politely not support.

I hope I'm not opening a can of worms with this question.

neilb:
I think we should allow / support people changing the case of their modules. Most of the examples I've seen have been internal modules, not really intended for direct public use. But I'm sure there will have been cases where people have changed the case of public-facing modules.

We already support changing case of modules in the indexer – the canonicaliser makes sure that all permissions are switched to the new casing. This works right now, I'm pretty sure, it's just historical cases that we're clearing up. The main thing is to ensure @wolfsage's change doesn't break that (I've not looked at it).

Yes this can cause problems, but it's a natural thing for people to do. For example, where there's a large project with multiple contributors, and someone adds a module that doesn't following their naming convention, so at some later date someone wants to bring it in line.

It's worth a blog post talking about the problems this can cause, so people are aware, and it would be even cooler if PAUSE could spot when this happens and direct people to the blog post (though most people don't read the emails from PAUSE, so *shrug*).

wolfsage:
I can't see how this change would cause that problem but I suggest we (I) write a test case if one doesn't already exist to account for this before merging.

andk:
Thanks, we let this be superceded by #440 but we cherry-picked the commit "mldistwatch: Rewrite perms/primeur in the transaction" from here. Deployed a few minutes ago.

PR #426, opened 2023-05-05
add a GitHub Actions workflow

rjbs:
If merged, this will mean that all pull requests will have the test suite run agains them.

Currently this skips the web tests because they require MySQL, which isn't set up.  If I *do* set them up, one of the tests seems to run forever.  I think we should run them all, but I don't know how to get there from here yet, and running *some* is a good start!


rjbs:
For an example of a passing test run, you can see https://github.com/rjbs/pause/actions/runs/4894217428

PR #430, opened 2023-05-25
Copyedit 04pause.html slightly

ap:
This replaces “matches X or matches Y” with “matches X or Y”.


andk:

PR #440, opened 2024-04-13
Add lc_package columns to avoid timeout/deadlock

charsbar:
This PR is another fix for #406. Contrary to #424, which removes `LOWER()` so that PAUSE can use indices (but lose case-insensitivity), this PR adds indexed lc_package columns to keep the current case-insensitive search and let PAUSE use indices at the cost of some extra space.

We may need to replace some more for the web app, but I haven't thoroughly checked yet. Also, we may need to add a unique lc_pack_user index (or replace pack_user with lc_pack_user), but I'm still wondering which is better.

If you try this PR before merging, apply one-off-utils/schemachange-2024-04.sql.

### DIFF
--- a/lib/PAUSE.pm
+++ b/lib/PAUSE.pm
@@ -345,7 +345,7 @@ sub owner_of_module {
                    FROM mods where modid = ?},
                  primeur => qq{SELECT package,
                           userid
-                   FROM primeur where LOWER(package) = LOWER(?)},
+                   FROM primeur where lc_package = LOWER(?)},
                 );
     for my $table (qw(mods primeur)) {
         my $owner = $dbh->selectrow_arrayref($query{$table}, undef, $m);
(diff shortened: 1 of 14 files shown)


lib/PAUSE.pm
```diff
@@ -345,7 +345,7 @@ sub owner_of_module {
                    FROM mods where modid = ?},
                  primeur => qq{SELECT package,
                           userid
-                   FROM primeur where LOWER(package) = LOWER(?)},
+                   FROM primeur where lc_package = LOWER(?)},
```
wolfsage, line 348: Any reason you left LOWER(?) in this one place but changed all others to use `lc`?

wolfsage (approved):
I think this is fine. I have two thoughts:

* If we used a case insensitive colation for lc_package, we wouldn't have to remember to compare with `lc $var` each time, which maybe provides a little less risk
* If we want to test this more before deploy we can stand this up on a DO box and do some manual testing (if we haven't already)

lib/PAUSE.pm
charsbar: Or it might be better to repeat selectrow_arrayref without the for loop.

wolfsage: Ah, I see. I think leaving LOWER(?) here is fine then

andk:

PR #441, opened 2024-04-13
Use Parallel::Runner to limit the number of mldistwatch processes

charsbar:
When you regenerate package-related tables from the existing CPAN directory to see how a patch for the indexer works, paused may spawn too many mldistwatch processes, which would cause various problems such as out-of-memory. This PR uses Parallel::Runner (by Exodist) to limit the maximum number of mldistwatch processes.


andk:
Thanks, applied and deployed about an hour ago

PR #445, opened 2024-04-25
cron/mysql-dump: add flexibility

rspier:
add --extended-insert flag to pass through to mysqldump to enable extended-insert (one big INSERT statement instead of one per row.) Set to 1 to enable.  

add --override-backupdir flag to override the output directory that is specified in the config.


rspier:
FYI @rjbs

Issue #436, opened 2024-03-21
The indexer needs to spot the new class syntax and create indexing permissions based on that


neilb:
The indexer looks for `package Foo::Bar` and then checks whether you (or anyone else) has an indexing permission on that package. And if no-one does, then you get first-come.

The new `class` feature introduces the `class Foo::Bar` syntax, and in code using this there is no `package` statement to spot.

In code which has `use feature 'class'` or `use Feature::Compat::Class`, then PAUSE should parse the class syntax. But maybe there are other modules which introduced `class` syntax but don't imply package?

This seems like a good topic for discussion at PTS 2024.

andk:
We believe this issue has been addressed by applying #447 which has been deployed earlier today

PR #447, opened 2024-04-25
Respect class/role if the PAUSE finds a few modules are used in the pacakge file

### LINKED ISSUE #390: Get a version from `(class|role) NAME VERSION` line
Since Perl 5.37.9, `class NAME VERSION` is a valid native class declaration. So it would be nice for PAUSE to get a version from the line. Also, `role NAME VERSION` would be available eventually (considering Object::Pad has that).

The change should be the following but we need to discuss what should be supported first, as it may be too early because other tools like ExtUtils::MakeMaker don't get a version from the line.

```
diff --git a/lib/PAUSE/pmfile.pm b/lib/PAUSE/pmfile.pm
index a637e04..374778e 100644
--- a/lib/PAUSE/pmfile.pm
+++ b/lib/PAUSE/pmfile.pm
@@ -227,7 +227,7 @@ sub packages_per_pmfile {
             $pline =~ m{
                       (.*)
                       (?<![*\$\\@%&]) # no sigils
-                      \bpackage\s+
+                      \b(?:package|class|role)\s+
                       ([\w\:\']+)
                       \s*
                       (?: $ | [\}\;] | \{ | \s+($version::STRICT) )
```

charsbar:
Fixes #390 and #436.

### DIFF
--- a/lib/PAUSE/pmfile.pm
+++ b/lib/PAUSE/pmfile.pm
@@ -238,7 +238,7 @@ sub packages_per_pmfile {
         }
 =cut
 
-        # use feature 'class'; enabels class (and role, though not implemented yet)
+        # use feature 'class'; enables class (and role, though not implemented yet)
         if ($pline =~ /^[\s\{;]*use\s+(?:feature|experimental)\s+[^;]+\b(?:class|all)[^;]*;/) {
             $package_or_class = 'package|class';
         }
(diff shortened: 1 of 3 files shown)


rjbs (approved):
This looks correct.  Adding a test should be pretty simple, so I think that you should add a test, or maybe a bunch of tests.  Still, approved nonetheless.

wolfsage (approved):

andk:
Thanks, applied and deployed just now!

Issue #442, opened 2024-04-25
add file on cpan listing authors with pumpking permission


haarg:
For MetaCPAN, we want to know the list of users that are authorized to release perl.

This is currently available at https://pause.perl.org/pause/authenquery?ACTION=who_pumpkin&OF=YAML, but it is an authenticated end point.

What would work better is if the list was included somewhere on CPAN. A `07pumpkings.txt` file or something similar.

charsbar:
@haarg, https://pause.perl.org/pause/query?ACTION=who_pumpkin should be public, though there might be something as I've got 500 when I visit there using MS Edge (though there's nothing wrong when I visit via curl or other browsers).

cf. https://github.com/andk/pause/blob/master/lib/pause_2017/PAUSE/Web/Config.pm#L89-L95

charsbar:
authenquery is for the authenticated user. https://pause.perl.org/pause/query?ACTION=who_pumpkin&OF=YAML should work for you, but I agree a static file would be convenient.

charsbar:
Confirmed 08pumpkings.txt.gz existed in my CPAN mirror. Closing.

PR #449, opened 2024-04-26
Add 08pumpkings.txt under CPAN/authors like 02authors.txt and such

charsbar:
Fixes https://github.com/andk/pause/issues/442 (hopefully)


andk:

PR #452, opened 2024-04-26
prep work for PAUSE upgrade

rjbs:
* eliminate some host-based config changing from the *last* upgrade; in the future, this kind of thing can be done with a changed PrivatePAUSE.pm
* chmod +x a cron job program
* improve the logger identification for default PAUSE code and for mldistwatch; more should be updated in the future
* teach the indexer to respect the `PAUSE.CLOSED` stopfile
* eliminate some unused Config settings
* stop updating the 07mirrors file


andk:

PR #454, opened 2024-04-26
cron/cron-daily.pl: if there is no 08pumpkings, do not fail

rjbs:
08pumpkins will not exist on the first run of the daily cron job.

That is okay!  Just treat it as if the file had been empty.


charsbar (approved):
Thanks. LGTM!

PR #455, opened 2024-04-26
Add MFA protection to several pages of the PAUSE

charsbar:


doc/authen_pause.schema.txt
```diff
@@ -56,6 +56,9 @@ CREATE TABLE usertable (
   `changed` int(11) DEFAULT NULL,
   changedby char(10) DEFAULT NULL,
   lastvisit datetime DEFAULT NULL,
+  mfa tinyint(1) DEFAULT 0,
+  mfa_secret32 varchar(16) DEFAULT NULL,
+  mfa_recovery_codes text DEFAULT NULL,
```
rjbs, line 61: I feel like I'm probably going to ask a *lot* of questions.  I hope it's useful and not just annoying.

1. Does `mfa` mean "MFA is enabled"?  If so, is there any value to having that *and* `mfa_secret32`?  I would expect that you can't temporarily turn off MFA and turn it back on with the same secret, so no extra field is needed.

rjbs: It would be good to store `mfa_last_verified_at` with the time window in which it was verified.  If a OTP can be used more than once, it isn't a OTP!  Better practice is something like:
* TOTP tokens change every 30s
* we accept the TOTP token for now, 30s ago, and 30s from now
* we store the time seed of any verified TOTP
* we never accept a TOTP with a seed less than or equal to the last one

Otherwise, the token used at 12:00:00 has 89 seconds to be exfiltrated and used again.

lib/PAUSE.pm
```diff
@@ -100,6 +100,7 @@ $PAUSE::Config ||=
      HTTP_ERRORLOG => '/usr/local/apache/logs/error_log', # harmless use in cron-daily
      INCOMING => $IS_PAUSE_US ? 'ftp://localhost/incoming/' : 'ftp://pause.perl.org/incoming/',
      INCOMING_LOC => '/home/ftp/incoming/',
+     INCOMING_TMP => '/tmp/',
```
rjbs, line 103: Why are we putting MFA-enabled-user uploads in a different place?

lib/pause_2017/PAUSE/Web/Plugin/WithMFAProtection.pm
```diff
@@ -0,0 +1,86 @@
...
+                    my $pause = $c->stash(".pause");
+                    my @recovery_codes = split / /, $u->{mfa_recovery_codes} // '';
+                    for my $code (@recovery_codes) {
+                        if (PAUSE::Crypt::password_verify($otp, $code)) {
+                            my $new_codes = join ' ', grep { $_ ne $code } @recovery_codes;
+                            my $dbh = $c->app->pause->authen_connect;
+                            my $tbl = $PAUSE::Config->{AUTHEN_USER_TABLE};
+                            my $sql = "UPDATE $tbl SET mfa_recovery_codes = ?, changed = ?, changedby = ? WHERE user = ?";
+                            $dbh->do($sql, undef, $new_codes, time, $pause->{User}{userid}, $u->{userid})
+                              or push @{$pause->{ERROR}}, sprintf(qq{Could not enter the data into the database: <i>%s</i>.},$dbh->errstr);
```
rjbs, line 36: Probably it's fine, but: probably we need to escape `$dbh->errstr` for entities?

lib/PAUSE.pm
charsbar: Isn't it confusing putting distributions that are not yet confirmed in the same /incoming/ directory? If not, I'm happy to put them under /incoming/ and remove some code.

doc/authen_pause.schema.txt
charsbar: > Does mfa mean "MFA is enabled"? If so, is there any value to having that and mfa_secret32

Maybe we don't need `mfa` column anymore.

lib/pause_2017/PAUSE/Web/Plugin/WithMFAProtection.pm
charsbar: As the PAUSE is now a Mojo app, error messages stored in the `ERROR` list should eventually be escaped at https://github.com/andk/pause/blob/master/lib/pause_2017/templates/layouts/layout.html.ep#L58 , but it's arguable if we should show this kind of raw error message to a user.

ap: This sort of internal error is really none of the user’s business. Normal users won’t be able to act on it anyway, so it’s no use to them, but it might be of use to an attacker, whom it might help with crafting an injection if such a vulnerability is ever present somewhere. Showing the raw error to the user is fine if the application is running in some kind of dev/debug mode, but in production it should only be logged somewhere (ideally with some kind of a request ID (which maybe already exists – apologies for not knowing)) and shown to the user only as a generic 500 “sorry, something went wrong” error screen. I don’t know how the rest of the code handles errors though (apologies again), so this comment might be about the codebase in general and not actionable within this PR.

lib/PAUSE.pm
rjbs: I hadn't gotten far enough, when reading this (commit by commit) to understand that the process was "upload, then confirm".  Having a second location makes sense.  (Obviously, it'd be great to not have to confirm after upload.)

rjbs:
I think I more or less understand this, but I have some questions about what our longer-term plan is or should be.  Rolling this out as a temporary security measure seems okay to me, but I think we'd get a lot more value out of real API tokens for uploading and for real sessions that can be re-authenticated.  I know that's not going to happen in the next two days, but maybe we can discuss approving that as the next goal tomorrow.

What I'd actually love to see is something like a paragraph explaining (to the users who will have to use this) how it's meant to work.  We'll need to show them something like this when it changes, and also it'd help me understand whether this is all going to work right!

lib/PAUSE.pm
charsbar: > Obviously, it'd be great to not have to confirm after upload

If you pass a valid otp field value when you upload a file, the otp validation is done at the time of upload. However, the valid otp may change while uploading a large file. So we need to give users a chance to try a new otp without uploading the same file again.

lib/pause_2017/PAUSE/Web/Context.pm
```diff
@@ -40,6 +41,17 @@ sub version {
   $version;
 }
 
+sub authenticator_for {
+  my ($self, $user) = @_;
+  my $cpan_alias = lc($user->{userid}) . '@cpan.org';
+  my $secret32   = $user->{mfa_secret32};
```
wolfsage, line 47: A comment here explaining that `$user->{mfa_secret32}` will be undef the first time this is set up, and that Auth::GoogleAuth will generate one for us, would be nice I think.

lib/pause_2017/PAUSE/Web/Controller/User/Mfa.pm
```diff
@@ -0,0 +1,80 @@
...
+  $pause->{mfa_qrcode} = $auth->qr_code;
+  if (!$u->{mfa_secret32}) {
+    my $dbh = $mgr->authen_connect;
+    my $tbl = $PAUSE::Config->{AUTHEN_USER_TABLE};
+    my $sql = "UPDATE $tbl SET mfa_secret32 = ?, changed = ?, changedby = ? WHERE user = ?";
+    $dbh->do($sql, undef, $auth->secret32, time, $pause->{User}{userid}, $u->{userid})
+      or push @{$pause->{ERROR}}, sprintf(qq{Could not enter the data into the database: <i>%s</i>.},$dbh->errstr);
+  }
+
+  if (uc $req->method eq 'POST' and $req->param("pause99_mfa_sub")) {
```
wolfsage, line 26: This block of code says:

1. The user is updating their mfa
2. If they provided a 6 digit code? Does it verify? No? Fail
3. Did they provide a backup code? Does it verify? No? Fail

This means that you can supply an empty string and successfully disable mfa

It would probably be better to invert the logic and start with `my $success = 0;` and then only set `$success` to true under very explicit circumstances. Then, we only take action if `$success` is true. This can help avoid these kind of logic bugs

lib/pause_2017/templates/user/mfa/edit.html.ep
```diff
@@ -0,0 +1,59 @@
+% layout 'layout';
+% my $pause = stash(".pause") || {};
+% my $cpan_alias = lc($pause->{HiddenUser}{userid}) . '@cpan.org';
```
wolfsage, line 3: I think `$cpan_alias` is unused in this file

lib/pause_2017/templates/user/mfa/edit.html.ep
```diff
@@ -0,0 +1,59 @@
...
+<img src="<%= $pause->{mfa_qrcode} %>">
+</div>
+% } else {
+<p>Submit 6-digit code to disable Multifactor Authentication.</p>
+<%= hidden_field "pause99_mfa_reset" => 1 %>
+% }
+
+<div>
+<p>CODE: <%= text_field "pause99_mfa_code" => '',
+    size => 10,
```
wolfsage, line 53: This is 1 short for recovery codes which means you can't use them to disable your mfa like `sub edit` intends in Mfa.pm

lib/pause_2017/PAUSE/Web/Context.pm
```diff
@@ -40,6 +41,17 @@ sub version {
   $version;
 }
 
+sub authenticator_for {
+  my ($self, $user) = @_;
+  my $cpan_alias = lc($user->{userid}) . '@cpan.org';
```
wolfsage, line 46: I don't think we should add the `@cpan.org` bit. I don't use a cpan email address and this confuses me. I would just have it be the PAUSE login name

lib/pause_2017/PAUSE/Web/Controller/User/Mfa.pm
```diff
@@ -0,0 +1,80 @@
...
+
+sub edit {
+  my $c = shift;
+  my $pause = $c->stash(".pause");
+  my $mgr = $c->app->pause;
+  my $req = $c->req;
+  my $u = $c->active_user_record;
+
+  my $auth = $c->app->pause->authenticator_for($u);
+  $pause->{mfa_qrcode} = $auth->qr_code;
```
wolfsage, line 17: Wow. I don't love this module. qr_code gives a link to quickchart.io which means the url containing the secrets to set up the 2fa are in someone else's web logs. I don't think we should do this.

I think we should construct our own images and inline them, something like:
```
  my $uri = "otpauth://totp/$label?secret=$key&issuer=$org.$domain";

  my $img = plot_qrcode($uri, {
    size          => 4,
    margin        => 4,
    version       => 1,
    level         => 'M',
    casesensitive => 1,
  });

  $img->write(data => \my $qr_png, type => "png")
    or die "Failed to write image: " . $img->errstr;

  my $data = URI->new("data:");
  $data->data($qr_png);
  $data->media_type('image/png');

  # And now in the html  <img src="$data" />
```

wolfsage:
I think there's a few real problems to fix, but otherwise this would be a good thing to do.

I have reservations about using Auth::GoogleAuth's qr_code image links.

I think we could relatively easily add a full set of tests to make sure the behaviour works as we expect so we don't have to hand test things every time. (You could inject a known totp secret into a test user, and either optionally provide a way (if tests don't already have it)) to fake the current time, or modify things so you can inject a timestamp into the totp stuff so we can have consistent tokens to test against)

lib/pause_2017/templates/user/mfa/edit.html.ep
rjbs: Is that true?
```
        my $code = encode_base32(urandom(6));
        $code =~ tr/lo/89/;
        $code =~ s/^(.{5})/$1-/;
        push @codes, $code;
```

So, 6 bytes b32 encoded is 10 bytes.  Then `l` and `o` become `8` and `9`, still 10 bytes.  Then we put a `-` in the middle.  Okay, you're right, thanks!

lib/pause_2017/PAUSE/Web/Context.pm
charsbar: The generated secret32 at https://metacpan.org/dist/Auth-GoogleAuth/source/lib/Auth/GoogleAuth.pm#L21-25 might not be strong enough. Is it better for us to generate it by ourselves using Crypt::URandom instead of just adding a comment?

lib/pause_2017/PAUSE/Web/Controller/User/Mfa.pm
charsbar: Inverted the logic with https://github.com/andk/pause/pull/455/commits/684f2f05c7474d86c4d31f7736ee9b0bcb712a60

lib/pause_2017/templates/user/mfa/edit.html.ep
charsbar: Removed $cpan_alias with https://github.com/andk/pause/pull/455/commits/ca5ea284250293120610be611bbf0302292ff933

lib/pause_2017/templates/user/mfa/edit.html.ep
charsbar: Fixed the size with https://github.com/andk/pause/pull/455/commits/efdf93115a3fec43c8291f39c4791e68c6689fc0

lib/pause_2017/PAUSE/Web/Context.pm
charsbar: Removed `@cpan.org` part with https://github.com/andk/pause/pull/455/commits/abb0141293fe982fd641110611e3da0fb04bca1a

lib/pause_2017/PAUSE/Web/Controller/User/Mfa.pm
charsbar: Add `_generate_qrcode` with https://github.com/andk/pause/pull/455/commits/c2472f1afce567909548f37d56d3f0d3cf4a74ec

charsbar: and modified parameters with https://github.com/andk/pause/pull/455/commits/0d09dbd3f79e836564bc85b971a97f2242bd5def

doc/authen_pause.schema.txt
charsbar: Removed mfa column with https://github.com/andk/pause/pull/455/commits/e1549a564d4e9ccd3aac35c2f730c14e71f77a7b

PR #457, opened 2024-04-26
Increase margins and paddings for some elements

charsbar:
Maybe adding a little more space would be better?

before:
before

after:
after


andk:

Issue #438, opened 2024-04-02
Add an admin interface for changing the account status, or at least to toggle between 'active' and 'nologin'


neilb:
As a PAUSE admin, I'm not able to change a user's state between `active` and `nologin` (the `ustatus` column in the DB) – I have to ask Andreas to do a DB update. It is possible to trigger a change to the state from nologin to active, by changing other fields in the account, but it would be helpful if I could explicitly change this field on a user.

charsbar:
An email to let the user to know onetime password should also be sent when the ustatus is change by an admin?

neilb:
> An email to let the user to know onetime password should also be sent when the ustatus is change by an admin?

My first gut reaction to this was "we can leave it to the admin to email, as they'll probably be replying", because that's what I do. But we should assume they'll remember, so probably yes, send a quick email.

neilb:
I've worked out that there's an interface-by-proxy, or whatever the right term is, at least for unblocking someone.

If I change the secret email address field, then if there's a nologin block on the user, it will be removed. I generally add "+pause" to the email address, and then immediately remove it.

There isn't any way for me to set nologin on a user, other than my existing UI, which is Andreas :-)

andk:
Thanks, Neil, for the request. The implementation in #458 has now been deployed

PR #458, opened 2024-04-27
Introduce change_user_status admin interface

charsbar:
Should fix https://github.com/andk/pause/issues/438


neilb:
Would be neat if this could be merged and deployed during or before PTS :-)

andk:

PR #459, opened 2024-04-27
remove a bunch of etc content that is no longer relevant

rjbs:
I'd like us to document "everything" in the repository, and one way to make that easier is to start by removing everything not in use any longer.


rjbs:

PR #461, opened 2024-04-27
PAUSE::Config: merge new values from bootstrap

rjbs:
A lot of things were configured in bootstrap because the primary PAUSE was running with the old filesystem layout.

I want to keep the default config matching the default behavior of both test and real PAUSE instances.  The bootstrap program should be able to build on default, rather than old settings from years ago.


rjbs:

rjbs:
Merged in b0a11970

PR #462, opened 2024-04-27
05news: add an announcement of the new PAUSE server

rjbs:


rjbs:
Has now been applied!

Issue #427, opened 2023-05-07
Delete-Files doesn't list files in the right order, if some releases used "v" in the version


neilb:
If an author uses "v" in the version of their distribution, for _some_ releases, then when you go to Delete Files, the files aren't listed in the right order.

Here's an example. The author started off using vanilla release numbers, then used "v" notation for a while (boo), and then dropped the "v" (yay):

![plosuring-delete-files](https://user-images.githubusercontent.com/478461/236674240-a837d74c-da62-4449-9b47-73d60323c72d.png)

The 0.2 and 0.3 files should appear at the end of the list, but they're in the middle.

Grinnz:
I expect this is a simple asciibetical filename sort so it looks expected to me. To sort the way suggested could be done via an extracted "distribution version", except that such a concept doesn't exist in PAUSE - but using https://metacpan.org/pod/CPAN::DistnameInfo or similar process to extract the filename version would probably be good enough.

charsbar:
Note: I implmented this in my local virtual machine. I dropped the current js sorter and sorted files on the serverside using new query parameters (sort_by and sort_order). It worked, until I pretended to be someone else. When I click a link to sort as someone else, pretension is gone. Adding parameters to pretend to the links to sort is bad.
Maybe I need to fix how to pretend before I revisit this.

ap:
> Adding parameters to pretend <s>to the links to sort</s> is bad.

FTFY 😉

In our application at $work, pretending to be someone else is done in the cookie. The `user_id` field of the cookie changes to the ID of the user you are pretending to be, and a `real_id` field is added with the ID of who you originally logged in as. If you log out while there is a `real_id` value in your cookie, you do not actually get logged out, it just swaps the value of `real_id` back into `user_id` and clears `real_id`. It is more or less like EUID/UID in Unix.

Issue #363, opened 2020-10-16
Deleting a lot of old releases of the same distribution is tedious


neilb:
This is feedback from Philip Brenan that I received via email.

> It is quite difficult to schedule the deletion of a lot of files because one has to delete three files for each upload.
> An option to delete a block of releases (say between one version number and another) might alleviate a lot of very fiddly clicking and, perhaps, encourage more people to delete files from PAUSE

karenetheridge:
Various scripts exist to automate this.  Mine is https://github.com/karenetheridge/misc/blob/master/install/generic/bin/pause-cleanup.

PR #463, opened 2024-04-27
Check related checkboxes

charsbar:
Should fix part of #363. This PR should allow people to check (at most) three related checkboxes with one check. Maybe we can go further if the issue noted in #427 is fixed.


charsbar:
It might be better to add the same code to the reindex page where tarballs and meta files are listed together, though I'm not sure if we need to show meta files in the first place.

andk:

PR #466, opened 2024-04-27
selfconfig: split rsyncd.conf back into etc

rjbs:
This keeps distinct files more clearly under source control.


rjbs:

PR #467, opened 2024-04-27
Use lc_package to join tables

charsbar:
lc_package columns were introduced by #440 , but they are not used in the webapp at all. At least we should use them when we join tables.


andk:
Thanks applied, and deployed

PR #468, opened 2024-04-27
No need to show .meta files in the reindex list

charsbar:


rjbs:
I think this looks good!

andk:

PR #470, opened 2024-04-27
raise max upload size

rspier:
nginx defaults to 1mb client_max_body_size which is very limiting for the large POSTs that PAUSE requires.

Raise to 250mb, which is a little larger than the largest files.

Note: an upload of a 200mb file failed, but in Plack/mojo, not nginx as before. An upload of 25mb (bigger than the average perl5 release) was fine.

### DIFF
--- a/bootstrap/selfconfig-root
+++ b/bootstrap/selfconfig-root
@@ -263,6 +263,8 @@ upstream pause {
 server {
   listen 80 default_server;
 
+  client_max_body_size 250m;
+
   location / {
      proxy_pass http://pause;
      proxy_set_header X-Forwarded-Host \$host;


charsbar:
It might be better to set MOJO_MAX_MESSAGE_SIZE to an appropriate size, too (cf. https://metacpan.org/pod/Mojo::Message#max_message_size)

charsbar:
Setting max_message_size doesn't seem to work for me. My local PAUSE accepts no matter how large a file is.

@rspier, do you remember what kind of failture you saw from Plack/mojo when you uploaded a 200mb file?

rspier:
It was a giant mojo stack trace.

Today, I can't replicate it.  (Yay?)  I was just able to upload 200MB of null bytes.

wolfsage:
This is actually controlled by this:
```
lib/pause_2017/PAUSE/Web.pm
14:  $app->max_request_size(0); # indefinite upload size
```

rspier:
So my stack trace must have been a fluke.

rspier:
The nginx change should be sufficient.

PR #472, opened 2024-04-28
doc: purge cruft and update install docs

rjbs:


rjbs:
@andk says he's happy with this.

PR #473, opened 2024-04-28
etc/rsyncd.conf: log to syslog (journald), not a file

rjbs:
This means we get the space management from the journal.


rjbs:
❗️ I already deployed this, this commit is just bringing the build in line with reality.

PR #475, opened 2024-04-28
paused: use the PAUSE::Logger to log

rjbs:
(while also logging to the PAUSE_LOG)


rjbs:
@andk says he's happy with this.

rjbs:
Applied.

PR #476, opened 2024-04-28
Mysql and logging fix

wolfsage:
Configure mysqld so utf8 data in fullname column behaves like it used to.

Also in bootstrap restart journald one last time so we actually get logs


rjbs (approved):

rjbs:
Applied!

PR #478, opened 2024-04-28
bootstrap: install crontab as a symlink

rjbs:
That way, repo updates update the crontab.


rjbs:
We agreed to not do this.

PR #479, opened 2024-04-28
Remove unused PAUSE 1999 version code

wolfsage:
This gets rid of all unused 1999 code as far as I can tell. There may be more things to cleanup but this is a good start.

It leaves some config stuff around (/tmp/pause_1999 is being used for some stuff on pause3) but I think that's okay and could be cleaned up later


wolfsage:
All tests pass. I think if we want to keep the old code around for reference we have a few good options:
 * Put it into attic/ (so it's clear this is retired code and we don't have to guess/remember)
 * Tag it, you can always check out the tag and look
 * Put it in a new repo pause-1999

wolfsage:
We're going to tag master, then merge the deletion

PR #481, opened 2024-04-28
use PAUSE::Logger in more Perl programs

rjbs:


rspier (approved):
Do you want to split out the script deletions into their own PR for clarity?  Probably not required, as they're all individual commits.

rjbs:
I probably should've done it that way, but I think it's okay.  Will rebase and merge soon.

PR #488, opened 2024-04-28
No need to mention perlmonks anymore as the incident happened years ago

charsbar:


ap:
That text comes from cdb5e7cde1c19f799cd57d1404eeeaa4700f37c5, which was in 2009. So even the “recently“ you left is well out of date. I suggest changing the wording to this:

````
"User '$u' set to nologin. Your account may have been included in a precautionary password reset in the wake of a data breach incident at some other site. Please talk to modules\@perl.org to find out how to proceed"
````

rspier:
I don't think we want this kind of thing to go to modules@ anymore, as it is public.  Password change things are kind sensitive.

ap:
Where would you suggest instead?

ap:
@rspier What contact would you suggest sending people to instead?

rjbs:
The PAUSE operating model says: 
> modules@perl.org is the public list. Any email sent to this address is forwarded to all of the PAUSE admins, and will appear in the public archive: http://www.nntp.perl.org/group/perl.modules/. Anyone can send email to this list. This is the official way to contact the PAUSE admins. Many of the admins are on IRC, but we prefer all requests go to the mail alias, so they're "on the record".

I don't have a strong feeling here, but I'd think we're okay sticking with that, then taking further communication private.

andk:

PR #489, opened 2024-04-30
fix quoting of email addresses

rjbs:
We saw cases where a user with the "real name" of `Bob"Smith` could cause problems in how we'd send them mail.  This branch will fix that.


rjbs:
The edit_cred pause_2017 test is failing.  I can't really tell why.  I'll look into it more sometime soon, but maybe somebody (charsbar… 😉) can tell.

I did edit `lib/pause_2017/PAUSE/Web/Controller/User/Cred.pm` so I will go read that first…

rjbs:
This may be ready for review…

andk:
Hi Ricardo,

what do you think about this pull request ATM? Could it be applied as it is? Would it be possible for you to resolve the four conflicts listed above?

charsbar:
cf. https://github.com/andk/pause/issues/537 which you might also want to address in this PR.

PR #490, opened 2024-05-02
crontab: quiet git commands

rspier:
Add the -q flag to the git commands so they only output if there's an error.

This will make the cron emails less frequent and hopefully more actionable.


rjbs (approved):
I like it. 👍🏽

PR #491, opened 2024-05-02
fix Use of uninitialized value warnings.

rspier:
Some fields in @$author end up being undefined.  Replace them with an empty string instead.

### DIFF
--- a/cron/cron-daily.pl
+++ b/cron/cron-daily.pl
@@ -715,6 +715,7 @@ sub authors {
       require Encode;
       for (@$author) {
         defined && /\P{ASCII}/ && Encode::_utf8_on($_);
+        $_ //= ""; # some fields are undef, make them empty instead.
       }
 
       # Surely no one has tabs in his or her email or full name, but let's


cron/cron-daily.pl
```diff
@@ -715,6 +715,7 @@ sub authors {
       require Encode;
       for (@$author) {
         defined && /\P{ASCII}/ && Encode::_utf8_on($_);
+        $_ //= ""; # some fields are undef, make them empty instead.
```
rjbs, line 718: Do all the warnings come from these lines below?

```perl
      s/\t/ /g for @$author[0,1,2];
      s/\t/%09/g for $author->[3];
```

That would imply undef userid, email, fullname, or homepage.

I looked at the database.  The only field ever null is `fullname`, and only five users have that.  I suspect it is no longer possible to achieve.  I think I'd rather we fix the data.  Then if it comes back, we know that we're allowing bogus data, but right now the oldest record like this is from 2008.

It's the work of 60s to fix the data in the database, updating five rows to set `fullname = ''`.  What do you think?

rspier:
That makes sense.

```
MariaDB [mod]> select * from users where fullname is null;
<snip for privacy>
+-----------+---------+---------------------+-------+-----------------+----------+----------+-----------+----------+------------+------------+-----------+
5 rows in set (0.007 sec)

MariaDB [mod]> update users set fullname=""  where fullname is null;
Query OK, 5 rows affected (0.011 sec)
Rows matched: 5  Changed: 5  Warnings: 0

MariaDB [mod]> select * from users where fullname is null;
Empty set (0.007 sec)

```

rspier:
cron-daily runs warning free now.

PR #493, opened 2024-05-03
Remove spurious log line adding to cron noise.

rspier:
We don't need the Info: line.  If it fails, it will error.  Otherwise we can assume it succeeded.


rjbs (approved):
👍🏽

PR #494, opened 2024-05-04
use $Logger in cronjobs

rjbs:
Now that the cron jobs are all(?) successfully loading PAUSE::Logger, we should use it.  This replaces "warn" calls with use of $Logger.

### DIFF
--- a/cron/cron-daily.pl
+++ b/cron/cron-daily.pl
@@ -80,7 +80,7 @@ sub read_errorlog {
     close LOG;
     report "\nErrorlog contains $errorlines lines today\n";
   } else {
-    warn "error opening $PAUSE::Config->{HTTP_ERRORLOG}: $!";
+    $Logger->log("error opening $PAUSE::Config->{HTTP_ERRORLOG}: $!");
   }
 }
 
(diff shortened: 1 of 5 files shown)


cron/cron-daily.pl
```diff
@@ -80,7 +80,7 @@ sub read_errorlog {
     close LOG;
     report "\nErrorlog contains $errorlines lines today\n";
   } else {
-    warn "error opening $PAUSE::Config->{HTTP_ERRORLOG}: $!";
+    $Logger->log("error opening $PAUSE::Config->{HTTP_ERRORLOG}: $!");
```
rspier, line 83: if this is an error, should it be fatal?

cron/cron-daily.pl
```diff
@@ -241,7 +241,8 @@ sub watch_files {
           quoteHighBit => 1,
         );
         my $v = $d->stringify($File::Find::name);
-        warn sprintf qq{Found a bad directory v[%s], rmtree-ing}, $v;
+
```
rspier, line 244: unnecessary blank line?

rspier (approved):
github's diff is not doing a great line displaying this change.

cron/cron-daily.pl
rjbs: I think the whole premise of this code is kinda wonky, so I'm gonna not touch it beyond changing its logging.  "Send me the day's web error logs" is odd.

cron/cron-daily.pl
rjbs: Most blanks are, right? :)

Anyway, so much of the PAUSE code is without any vertical whitespace.  Sometimes I add some to help break things up.  This is maybe not the most obvious place to put it, but: I will continue to add vertical whitespace to create paragraphs of code.

rspier: I like vertical whitespace, this just felt like a weird place to put it.

PR #498, opened 2024-05-04
reduce mldistwatch log noise, and use constants (for errors

rspier:
Primarily

    Use a set of constant enums for when mldistwatch decides to ignore a dist.
    
    - The actual values are the same, where possible.  Some values were just '1'
      which has been made more specific.
    - These are only used in one place.
    - Now we can easily route the log lines for metadata and readme files to the
      debug log

But also a few other cleanups.


rspier:
Mostly obsoleted by #499 and #501 .

PR #500, opened 2024-05-05
Explicitly set a default branch when initializing the test git repo.

rspier:
This avoid git printng out a ~10 line warning, which clutters the output.

Do not merge until Perl/docker-perl-tester#62 is merged and deployed.  These flags do not work on buster, so we need a bookworm image to run on.


rspier:
Likely obsoleted by #501

PR #501, opened 2024-05-05
Run tests on bookworm, not buster.

rspier:
Bookworm matches what's in production. Buster is `oldoldstable`


rjbs (commented):
Good idea. LGTM.

PR #502, opened 2024-05-05
webui: eliminate descriptions of SQL in web interface

rjbs:
I think it's weird for the UI to describe the tables it will query. Users will want to know what information they're getting, but not how it's laid out in the database.


charsbar:
The changes for admin/pumpkin.html look good to me. However, as for the changes for (dist)perms/peek.html, I wonder if we need to remove the second and third lines. They may be apparent, but they are helpful explanations.

andk:
Thanks, #560 is applied and deployed

PR #505, opened 2024-05-06
Remove indexscripts cronjob and mentions.

rspier:
It has not been touched maintained in over a decade. For the past few years, at most 1 .pl file a year has been uploaded and indexed.  The new way is to use the App:: namespace. It has hardcoded paths to /home/kstar, which no longer exists.

Also note that lib/PAUSE.pod is also 10+ years out of date.


rjbs (approved):
👍🏽  Obviously, gotta get announced on pause-ops.  The "only one .pl file uploaded" in  years is good intel.

I think we should probably remove PAUSE.pod entirely, but this is a good start.

kstarsinic:
I, kstar, endorse this change. :)

rspier:
_phew_

PR #506, opened 2024-05-06
One-line patch to remove an unused variable in PAUSE::may_overwrite_file

ap:


rjbs:
I will check this out after a larger branch that touches some of this code lands.  I'm sure this is fine, I just want to avoid a merge conflict.

ap:
Has the larger branch landed?

ap:
Ping?

This patch does nothing other than remove a narrowly scoped lexical variable that is referenced only in its declaration.

PR #507, opened 2024-05-09
Minor doc updates and bootstrap cleanups

rspier:


bootstrap/selfconfig-root
```diff
@@ -71,6 +71,7 @@ my @required_debs = qw(
  libssl-dev
  nginx
  python3-certbot-nginx
+ rename
```
rjbs, line 75: Okay, but out of curiosity where is it used?

rjbs (approved):
👍🏽

bootstrap/selfconfig-root
rspier: It's used by journal-length-exporter, which is non-important monitoring, so isn't part of the normal bootstrap.  Maybe it shouldn't be here?

rspier: I dropped it for now. It's not actually needed by pause.  (There is still an open question of how we want to track monitoring.)

wolfsage (approved):

PR #508, opened 2024-05-09
GitHub Action improvements, including coverage tests!

rspier:
- Various cleanups of obsolete things from the workflow.
- Caching so we don't have to install modules each time (saves 2 minutes)
- Run Devel::Cover coverage tests and push results to Coveralls if the want-coverage label is applied.
- Explicitly use MariaDB, instead of whatever Debian's default is.

### DIFF
--- a/.github/workflows/perl-tests.yml
+++ b/.github/workflows/perl-tests.yml
@@ -8,6 +8,15 @@ on:
     - reopened
     - labeled
 
+#
+# * To trigger a coverage run, add the `want-coverage` label, and
+# cover will be used instead of prove.  Results will be posted to
+# https://coveralls.io/github/andk/pause
+#
+# * To flush the GitHub Actions Cache:
+# https://github.com/andk/pause/actions/caches
+#
+
 jobs:
   the-tests:
     runs-on: ubuntu-latest
--- a/.github/workflows/perl-tests.yml
+++ b/.github/workflows/perl-tests.yml
@@ -1,5 +1,12 @@
 name: "perl test suite"
-on: [ push, pull_request ]
+on:
+  push:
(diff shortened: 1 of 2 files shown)


.github/workflows/perl-tests.yml
```diff
@@ -30,13 +30,24 @@ jobs:
...
+            /usr/local/lib/perl5/site_perl
+          key: ${{ runner.os }}-${{ steps.get-perl-version.outputs.version }}-${{ hashFiles('Makefile.PL') }}
       - name: Install prereqs (cpan)
         # This could probably be made more efficient by looking at what it's
         # installing via cpanm that could, instead, be installed from apt.  I
         # may do that later, but for now, it's fine! -- rjbs, 2023-01-07
         run: cpanm --notest --installdeps .
       - name: Run the tests
-        if: !(contains(github.event.pull_request.labels.*.name, 'want-coverage'))
+        if: "!(contains(github.event.pull_request.labels.*.name, 'want-coverage'))"
```
rjbs, line 54: does this belong in its own commit?  I think yes: not a showstopper but also easy to change

rjbs (approved):
Looks good.  Took me a bit to make sure I understood it.  I noted one very very small nit in git discipline.

.github/workflows/perl-tests.yml
rspier: Good catch.  It actually belongs in the first commit, not this one.

rspier: (i.e. it was a yaml format error without the quotes, and apparently I didn't add it to the right one the first time.  It is now part of the first commit that adds coverage.)

PR #512, opened 2024-05-11
improve how we track updating of 02packages after test indexing

rjbs:
This replaces the generic methods for checking file changes.  They were generic (great!) but also unreliable and kind of a pain to call.  The new method is more reliable and easier to use.


rjbs:
I've rebased #510 onto this work to keep life simple.

PR #515, opened 2024-05-12
GitHub Action improvements, including coverage tests! 

rspier:
Various cleanups of obsolete things from the workflow.
Caching so we don't have to install modules each time (saves 2 minutes)
Run Devel::Cover coverage tests and push results to Coveralls if the want-coverage label is applied.
Explicitly use MariaDB, instead of whatever Debian's default is.

Same as #508, but going to merge without rebase.

### DIFF
--- a/.github/workflows/perl-tests.yml
+++ b/.github/workflows/perl-tests.yml
@@ -8,6 +8,15 @@ on:
     - reopened
     - labeled
 
+#
+# * To trigger a coverage run, add the `want-coverage` label, and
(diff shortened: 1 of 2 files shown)


rspier:
This is identical to #508, which was approved by @rjbs, but I merged that with rebase, and this was big enough that it should have a merge commit to track it.  So I did some things I don't want to mention to make that happen.  Hopefully everyone else was asleep and didn't notice.  :wink:

PR #516, opened 2024-05-12
Workflow to implement a "do not merge" label.

rspier:
From https://github.com/jessesquires/gh-workflows/blob/main/.github/workflows/do-not-merge.yml (MIT licensed)

### DIFF
--- a/.github/workflows/do-not-merge.yml
+++ b/.github/workflows/do-not-merge.yml
@@ -1,0 +1,17 @@
+name: Do Not Merge
+
+on:
+  pull_request:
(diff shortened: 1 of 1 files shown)


rjbs:
How does this work?  Is the idea that we have configured the repository to refuse merge when any check fails, or is there more to it than this?

rspier:
> How does this work? Is the idea that we have configured the repository to refuse merge when any check fails, or is there more to it than this?

It's exactly that.  The repository is not currently configured that way -- I don't have permission to do so.  But once the check is in place... it could be.

rjbs (approved):

PR #519, opened 2024-05-15
tests: switch to slim-bookworm base image

rspier:
The slim-bookworm base image (https://github.com/Perl/docker-perl-tester/pull/66) shrinks the image size from 1.6GB to 570GB, and reduces container initialization time by 80%.  (from 45sec to 12sec.)

Because the slim container doesn't have all the batteries included, we need to add some more packages to facilitate a few modules:
gnupg for Module::Signature
libdb-dev for DB_File
libexpat-dev for XML::parser
zip for unzip
nodejs for actions/cache

### DIFF
--- a/.github/workflows/perl-tests.yml
+++ b/.github/workflows/perl-tests.yml
@@ -33,7 +33,16 @@ jobs:
       - name: Install prereqs (apt)
         run: |
           apt update
-          apt install -y rsync mariadb-server
+          apt install -y \
+            gnupg \
+            libdb-dev \
+            libexpat-dev \
+            libmariadb-dev-compat \
+            libxml2-dev \
+            mariadb-server \
+            nodejs \
+            rsync \
+            zip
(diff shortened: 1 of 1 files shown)


.github/workflows/perl-tests.yml
```diff
@@ -33,7 +33,7 @@ jobs:
       - name: Install prereqs (apt)
         run: |
           apt update
-          apt install -y \
+          apt-get install -y \
```
rjbs, line 36: typo in commit "stable interfface" should only have one "f"

rjbs (approved):
👍🏽 (but typo in commit you should fix)

wolfsage:
> Use the slim-bookworm base image, it's 1/3 the size!

If this commit brings the time down to run the test suite significantly, I think that should be part of the commit.

"It's 1/3 the size" doesn't really tell us why the commit exists, and commit messages should explain their purpose I think

PR #520, opened 2024-05-16
tests: save plack logs to a file.

rspier:
Save the plack logs from tests to a file, and then upload that file as an artifact. This significantly quiets/shortens the output from the tests.

Implement by creating a new plack log environment config `ghaction`.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -1,0 +1,17 @@
+# GitHub Actions Workflows
+
+This directory contains workflow definitions for GitHub actions.
+
+## Local Testing
+
+The easiest way to end to end test them is wih a tool such as
+https://github.com/nektos/act (https://nektosact.com/).
+
+```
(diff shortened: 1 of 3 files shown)


rspier:
@charsbar WrapAction.pm line 28 is one of the remaining sources of logs during the tests.  I played with changing it, but only ended up breaking other tests.  (Weird that changing something written to STDERR could break tests....)  Any thoughts on the best way to send those errors through the logging framework?

I know @rjbs is thinking about long-term logging stuff, but in the short- term I'm just trying to improve developer experiences.

rjbs (approved):
I am not a GH expert or a Mojo expert, but I this seems sensible from what I understand.

rspier:
@charsbar I've merged this because it's definitely a dev usability improvement.  I'm still interested in improving WrapAction's output, and to know if this was the best way to do it.

PR #522, opened 2024-05-18
Log error messages via $c->app->pause->log instead of STDERR

charsbar:
cf. #520 @rjbs , @rspier


rspier:
That is beautiful!  Thank you!

rspier (approved):

rspier:
Rebased against master before merging.  Thanks for helping reduce the noise.

PR #524, opened 2024-05-19
Fix warning about spurt

rspier:
Mojo::File::spurt is deprecated in favor of Mojo::File::spew at t/pause_2017/action/add_uri.t line 37.

### DIFF
--- a/t/pause_2017/action/add_uri.t
+++ b/t/pause_2017/action/add_uri.t
@@ -34,7 +34,7 @@ subtest 'get: user with subdirs' => sub {
         my $user_home = $PAUSE::Config->{MLROOT}."/".PAUSE::user2dir($user);
         my $subdir = path("$user_home/test");
         $subdir->make_path;
-        $subdir->child("stuff.txt")->spurt("Foo");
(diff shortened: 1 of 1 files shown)


charsbar:
This is only acceptable when the version of Mojolicious in the production server is upgraded to the version 9.34 or higher. 
cf. https://metacpan.org/dist/Mojolicious/changes#L18 and https://metacpan.org/dist/Mojolicious/changes#L22

rspier:
Looks OK to me.

```pause@pause3:~$ which perl
/home/pause/.plenv/shims/perl
pause@pause3:~$ perl -MMojolicious -le'print $Mojolicious::VERSION'
9.36
```

rspier:
@rjbs Do we want to pin all the versions so they line up between test and prod to try and prevent surprises?

wolfsage:
> @rjbs Do we want to pin all the versions so they line up between test and prod to try and prevent surprises?

IMO yeah, our builds should give us the same setup as prod

rjbs:
@rspier Yes.

rspier:
tracking that in #529

PR #530, opened 2024-05-20
Setup enforce-rebase action

rspier:
Based on https://github.com/cyberark/enforce-rebase

### DIFF
--- a/.github/workflows/enforce-rebase.yml
+++ b/.github/workflows/enforce-rebase.yml
@@ -1,0 +1,24 @@
+# This workflow will fail if the branch contains a merge commit or if
+# it isn't based on master.
+#
+# Caveat: It runs at push time, so if master moves after the branch is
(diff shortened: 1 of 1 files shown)


rspier:
You can see my noodling with this including passes and fails at https://github.com/rspier/pause/actions?query=enforce-rebase

haarg:
GitHub's branch protection rules have a "Require branches to be up to date before merging" setting (under "Require status checks to pass before merging") that may help with this.

rspier:
Thanks Graham.  I wonder if that's new.  It's definitely a bit buried.  And that's definitely cleaner than a custom workflow.

image

We might actually want *both* this check and GitHub's enforcement.

Here's what it looks like in action.
image

Someone inadvertantly clicking Update Branch could create the merge commit we're trying to avoid.


@andk is the only one with admin permission to set it up right now.

FYI @rjbs

PR #532, opened 2024-05-23
actions: Remove upload-artifact@v3 hack.

rspier:
Act now supports upload-artifact@v4, so we don't need to special case that in testing.


rspier:
Successfully tested with:
```
../act/act --artifact-server-path=/tmp/artifacts
```

act repo is at e1e5671e3d9e (committed May 20).

rjbs (approved):
👍🏽 Looks good.

PR #543, opened 2024-12-21
bootstrap-root: Cleanups for a lighter weight Docker environment.

rspier:
- don't install reccomended packages
- support not configuring certbot
- make firewall setup optional
- don't recreate users if they already exist
- chown the whole data directory contents, not just the directory itself
- don't recreate the homedir if it already exists


rspier:
I found this sitting around since last year.  (Oops!)

andk:
Thanks, applied

PR #549, opened 2025-03-31
Updated the process for what to do when a CPAN author passes away

neilb:
This updates the section in the Operating Model which describes what to do when an author passes away. Over time I've gradually refined this process and have been running off my own checklist. With the recent passing of ZEFRAM, we also identified some additional steps.


andk:
Thanks, applied  and deployed just now

PR #550, opened 2025-04-10
GET URL: HTTPS is now supported

thibaultduponchelle:
# Description 

In short, GET URL supports HTTPS

The note added in commit [a1022e72ae373a7a593317d4d643a58c611d41f0](https://github.com/andk/pause/commit/a1022e72ae373a7a593317d4d643a58c611d41f0) one year ago is no longer true

pause-get-url-https

### DIFF
--- a/lib/pause_2017/templates/user/uri/add.html.ep
+++ b/lib/pause_2017/templates/user/uri/add.html.ep
@@ -56,8 +56,7 @@ you do not seem to want HTTP upload enabled, we do
 % }
 
 <p class="url_upload"><b>GET URL:</b> PAUSE fetches any http or ftp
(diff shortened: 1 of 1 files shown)


wolfsage:
Thanks, could you update the commit message to explain a little bit more? "GET URL: HTTPS is now supported" doesn't tell us when or why that is true, which leaves the reader with questions.

I suggest:

```
GET URL: HTTPS works again

It stopped working for awhile when the PAUSE machine was too old to get OpenSSL updates, but around April of last year PAUSE was migrated to a newer OS and since then HTTPS has been working fine
```

thibaultduponchelle:
> [...] could you update the commit message [...]

I updated the commit message with your recommendations, thank you 😃

PR #551, opened 2025-04-15
Update PAUSE Operating Model for giving up first-come permissions

thibaultduponchelle:
# Update PAUSE Operating Model documentation
The current operating model is outdated and says that we can drop first-come without transferring to anybody (which implies to release the namespace to the public) whereas in practice the first-come is transferred to ADOPTME.


charsbar:
cf. https://github.com/andk/pause/issues/552

andk:
Thanks, applied (with a small tweak) and deployed just now

PR #559, opened 2025-05-03
Set content_disposition and content_type headers before serving a generated YAML file

### LINKED ISSUE #545: YAML download for "View Permissions" results does not suggest filename
When downloading the YAML file linked at the bottom of the "View Permissions" page, the filename is *authenquery*. The content type is `application/octet-stream` and no filename is suggested. Here's the response header:

```
$ curl "https://${PAUSEID}:${PAUSEPASS}@pause.perl.org/pause/authenquery?ACTION=peek_dist_perms&OF=YAML&pause99_peek_dist_perms_sub=1&pause99_peek_dist_perms_by=a&pause_peek_dist_perms_query=ADOPTME" -v
* Host pause.perl.org:443 was resolved.

... lots of stuff ...

* Request completely sent off
< HTTP/1.1 200 OK
< Server: nginx/1.22.1
< Date: Wed, 01 Jan 2025 17:28:13 GMT
< Content-Type: application/octet-stream
< Content-Length: 5956
< Connection: keep-alive
< Vary: Accept-Encoding
< Pragma: no-cache
< Cache-Control: no-cache
```

YAML now has `application/yaml` as defined by [RFC 9512](https://www.rfc-editor.org/rfc/rfc9512.html), and there are various tricks to suggesting the filename (although I wouldn't try that hard to please all browsers).

---

As a bonus, I found [yaml2json](https://github.com/bronze1man/yaml2json/releases):

```
$ curl ... | yaml2json | jq ...
```

charsbar:
should fix #545

### DIFF
--- a/lib/pause_2017/PAUSE/Web/Plugin/RenderYAML.pm
+++ b/lib/pause_2017/PAUSE/Web/Plugin/RenderYAML.pm
@@ -13,6 +13,7 @@ sub register {
     my $dump = YAML::Syck::Dump($data);
     my $edump = Encode::encode_utf8($dump);
     my $action = $c->req->param('ACTION') || 'pause';
+    $action =~ tr/a-z0-9_//cd;
     $c->res->headers->content_disposition("attachment; filename=$action.yaml");
     $c->res->headers->content_type('application/yaml');
     $c->stash(format => "text");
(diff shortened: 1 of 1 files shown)


lib/pause_2017/PAUSE/Web/Plugin/RenderYAML.pm
```diff
@@ -12,6 +12,9 @@ sub register {
     local $YAML::Syck::ImplicitUnicode = 1;
     my $dump = YAML::Syck::Dump($data);
     my $edump = Encode::encode_utf8($dump);
+    my $action = $c->req->param('ACTION') || 'pause';
+    $c->res->headers->content_disposition("attachment; filename=$action.yaml");
```
Grinnz, line 16: Mojo won't allow to set newlines in the header value, but I'd still be careful about directly injecting a request parameter into a response header.

andk:

PR #560, opened 2025-05-03
Eliminate descriptions of SQL in web interface

charsbar:
the same as https://github.com/andk/pause/pull/502 , but restored statements we should not remove


andk:
Thanks, deployed just now

PR #561, opened 2025-05-03
Improve sorting by the modified date

### LINKED ISSUE #553: Sort on 'Modified' column is 'Show my files' page is lexicographic, rather than by date
Howdy,

The dates in the 'Modified' column look like

`Wed, 31 May 2017 19:57:53 GMT`

So when they are sorted in reverse order, entries prefixed with 'Wed' show up first, with 'Fri' entries being last. Vice versa for normal order.

My expectation is that they would sort according to

  YYYY-MM-DDTHH:MM:SS

instead.

Cheers!
Diab

charsbar:
Store raw modified time in a "data-modified" attribute of each "modified" cell and configure list.js to sort by the value, instead of sorting by the blurb value.

should fix #553 .


andk:

PR #562, opened 2025-05-03
Use distvname instead of dist not to check distributions with different versions

charsbar:
Sorry, #463 was only good if there are no distributions with different versions. Using distvname instead of dist should fix the issue.


andk:

PR #563, opened 2025-05-03
Skip git manipulation if GITROOT is empty

charsbar:
Keeping track of the changes done on the production environment is nice, but it is useless while developing or testing on a local environment. It would be nice if we could skip git manipulation by setting GITROOT to empty, or by any other mean.


andk:

PR #567, opened 2025-05-13
Revert changes done for show_files/delete_files at PTS 2025

### LINKED ISSUE #566: Deleting files without a distv is not possible anymore
@charsbar @andk : If I go to "Delete Files" and select _one_ checkbox for an upload which is not a normal perl distribution (e.g. a patch file), then *all other non-distribution files* are also selected. In the screenshot below I just clicked on one of the patch/* files:

Image

The HTML for one entry looks like this:
```
<input data-distv="" name="pause99_delete_files_FILE" type="checkbox" value="patches/GD-2.56-RT114788-GREGOA.patch">
```
That is, `data-distv` is empty, which is probably causing the problem.

Maybe introduced by this change?
https://github.com/andk/pause/commit/08e28db4e3c5813ff476a99f853c26d984abafb1

charsbar:
This reverts #562, #561, and #463, and should fix #566.

@andk, sorry for the trouble. Could you review this and merge/deploy if it's ok?

### DIFF
--- a/lib/pause_2017/PAUSE/Web/Controller/User/Files.pm
+++ b/lib/pause_2017/PAUSE/Web/Controller/User/Files.pm
@@ -4,7 +4,6 @@ use Mojo::Base "Mojolicious::Controller";
 use HTTP::Date ();
 use File::pushd;
 use PAUSE ();
-use CPAN::DistnameInfo;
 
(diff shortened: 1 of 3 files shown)


andk:

PR #570, opened 2025-07-13
Switch to object interface for creating tempdirs

### LINKED ISSUE #533: _rewrite_tarball in paused should remove the tmpdir after use

rspier:
Switch to the object interface for creating temp directories.  This primarily changes their lifetime.  They are now cleaned up when the scope ends, as opposed to when the program exits.

Fixes #533


andk:
Thanks, applied and deployed

PR #572, opened 2025-09-11
add reference to modules@perl.org

karenetheridge:
re https://superuser.com/questions/1922570/perl-module-publishing-error-on-web-login-attempt-to-pause-and-additionally-ca and https://www.nntp.perl.org/group/perl.modules/2025/08/msg106169.html

### DIFF
--- a/htdocs/04pause.html
+++ b/htdocs/04pause.html
@@ -99,6 +99,12 @@ etc.</p>
 href="http://pause.perl.org/pause/query?ACTION=mailpw"
 >Non-SSL version</a>) for further instructions.</p>
 
+<p>If you're still having difficulties, please send an email to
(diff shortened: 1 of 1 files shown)


ap:
Links labelled “here” are bad style for various reasons, one of them being accessibility. (A good way of chosing good link labels is to imagine all other text being gone: the labels should still make some kind of sense without supporting context. Screen readers typically offer such a mode where they only read out the links, for example, and “here; here; here; here” is not helpful.)

I suggest something like
````html
It has <a href="https://lists.perl.org/list/modules.html">public archives</a>,
so don't include any confidential information in your message.</p>
````

karenetheridge:
thanks for the feedback; updated.

andk:
Thanks, applied, and  deployed just now

PR #573, opened 2026-04-12
Speed up query in paused loop.

rspier:
- Add an index on dverified.
- Stop doing explicit conversion from str to int (dverified + 0)
- dverified will always be a 10 digit unix timestamp, so string comparison results in the same as numeric.
- Query goes from ~30 sec to ~5

(Alternative considered was adding a virtual dverified_num column, but that's a lot more invasive.)

### DIFF
--- a/bin/paused
+++ b/bin/paused
@@ -201,7 +201,7 @@ sub loop () { # we're NOT called as a method
 	      WHERE length(userid) > 1
                 AND ( dverified=''
                       OR
-                      dverified+0 > ?
+                      dverified > ?
                     ) }; #};
 
   $sth = $dbh->prepare($query);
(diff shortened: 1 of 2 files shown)


andk:
The reason why we are using the +0 are all the dates in the table with 9 digits.

```
MariaDB [mod]> with cte as ( select dverified+0>1000000000 dv from uris ) select dv, count(*) cnt from cte group by dv;
+----+--------+
| dv | cnt    |
+----+--------+
|  0 |  15189 |
|  1 | 395436 |
+----+--------+
2 rows in set (3.509 sec)
MariaDB [mod]> select max(dverified) from uris where dverified > '1775030167' ;
+----------------+
| max(dverified) |
+----------------+
| 999996699      |
+----------------+
1 row in set (0.001 sec)

MariaDB [mod]> select max(dverified) from uris where dverified+0 > '1775030167' ;
+----------------+
| max(dverified) |
+----------------+
| 1776028378     |
+----------------+
1 row in set (0.915 sec)
```

Did you consider that?

rspier:
Totally missed that.  I'll consider re-spinning this as a virtual column, but not sure when I'm going to get to it.  It almost might complicate the SQLite fallback.  Really, this column should be an INT.

PR #575, opened 2026-04-24
Make it easy to extend pause web

charsbar:
This PR is intended to make it easier to extend the current PAUSE Web UI.

With this, we can add files such as `lib/pause_2026/PAUSE/Web2026/Controller/ForNewFeature.pm` and `lib/pause_2026/templates/for_new_feature.html.ep`  without modifying files under `lib/pause_2017` (or adding new files under 'lib/pause_2017` ) .

Also, this fixes the wrong usage of `$app->secrets`.


andk:
And deployed

PR #578, opened 2026-04-25
Replace Path::Class used in tests with Path::Tiny

### LINKED ISSUE #526: replace Path::Class in test suite with Path::Tiny
…and when doing that, replace `filestate` in TestPAUSE with a digest check.

charsbar:
Should fix #526, though I haven't fixed Path::Class usage in lib/PAUSE/mldistwatch.pm and one-off-utils/build-fresh-cpan , and thus I haven't updated Makefile.PL and cpanfile yet.


andk:
Thanks, applied and deployed

PR #577, opened 2026-04-25
Rename who_pumpkin as perl_releasers

charsbar:
Should fix #535 . See also #534 , but I haven't modified the operating model yet.

PR #579, opened 2026-04-25
Add a hidden redirection from who_pumpkin to perl_releasers

charsbar:
To address https://github.com/andk/pause/pull/577#issuecomment-4318623247 .


ap:
Arigatou 🙂

andk:
Thanks, applied and deployed

PR #581, opened 2026-04-25
Remove cpan.org email forwarding options and add shutdown/cpan-security notices

rspier:
I was unable to test this locally, because I don't have a great pause dev environment.  It intentionally just modifies the frontend.


domm:
@rspier I have made pause (hopefully) easy runable via docker-compose, so you might now be able to test this locally :-)

PR #585, opened 2026-04-26
Add a temporary warning

### LINKED ISSUE #584: How to deal with the E-Mail forwarding service being in limbo
Tentative collection of measures we need to get into effect today (day one without email forwarding)
 
- global idea on this changeset
  - introduce a global variable $cpan_dot_org_mail_forwarding_ok = 0;
  - all parts of the user interface that rely on a working
    cpan.org-mail-forwarding-service must be modified in a way that
    makes it clear to the user that the service is at the moment not
    available. Bike shedding follows:
    - text referring to these features should get a consistent markup:
      maybe start the text parts with a not too intrusive "under
      construction" symbol and change the foreground and background
      color to a unique style not used elsewhere in the user interface
    - any markup is allowed here as long as it is not misleading the
      user
  - the resulting code should be switchable back to the currently
    visible userinterface as soon as we change the global variable to a
    true value.
- concrete changes
  - (conditionally on $cpan_dot_org_mail_forwarding_ok = 0) before the
    sentence "The email address .... should be configured to forward"
    insert in bold "🚧 The mail forwarding service is currently not
    available. Changes here will only take effect when the service can
    be started up again. -- "
  - apply these two hunks of pull request #581 unconditionally:

```
    diff --git a/lib/pause_2017/templates/admin/email_for_admin.html.ep b/lib/pause_2017/templates/admin/email_for_admin.html.ep
    index d1dad403..8b6abc84 100644
    --- a/lib/pause_2017/templates/admin/email_for_admin.html.ep
    +++ b/lib/pause_2017/templates/admin/email_for_admin.html.ep
    @@ -9,7 +9,7 @@
       <thead>
         <tr>
           <th class="sort asc" data-sort="userid">id</th>
    -      <th class="sort" data-sort="email"><i>id</i>@cpan.org gets forwarded to</th>
    +      <th class="sort" data-sort="email">Contact Email</th>
         </tr>
       </thead>
       <tbody class="list">
    diff --git a/lib/pause_2017/templates/public/request_id/_form.html.ep b/lib/pause_2017/templates/public/request_id/_form.html.ep
    index b090853b..86d85028 100644
    --- a/lib/pause_2017/templates/public/request_id/_form.html.ep
    +++ b/lib/pause_2017/templates/public/request_id/_form.html.ep
    @@ -23,6 +23,7 @@ Note: You can enter fairly free-form text here but it must consist of at least t
     <p>
     <%= text_field "pause99_request_id_email", size => 32 %>
     </p>
    +<p><small>Note that contact information will be shared with the CPAN security team (<a href="https://security.cpan.org/">security.cpan.org</a>).</small></p>
     </div>
     
     <div class="alternate<%= $alt++ % 2 + 1 %>">
```

  - (conditionally on $cpan_dot_org_mail_forwarding_ok = 0) add the paragraph:
```
    <p><small>Note that contact information will be shared with the E-Mail forwarding service.</small></p>
```

charsbar:
This PR should fix #584 .

### DIFF
--- a/lib/PAUSE.pm
+++ b/lib/PAUSE.pm
@@ -27,6 +27,9 @@ use Sys::Hostname ();
 use Time::Piece;
 use YAML::Syck;
 
+# cf. https://github.com/andk/pause/issues/584
+our $cpan_dot_org_mail_forwarding_ok = 0;
+
 our $USE_RECENTFILE_HOOKS = Sys::Hostname::hostname =~ /pause/;
 if ($USE_RECENTFILE_HOOKS) {
   unless ($HAVE_RECENTFILE) {
--- a/lib/pause_2017/templates/admin/email_for_admin.html.ep
+++ b/lib/pause_2017/templates/admin/email_for_admin.html.ep
@@ -9,7 +9,7 @@
   <thead>
     <tr>
       <th class="sort asc" data-sort="userid">id</th>
-      <th class="sort" data-sort="email"><i>id</i>@cpan.org gets forwarded to</th>
+      <th class="sort" data-sort="email">Contact Email</th>
     </tr>
   </thead>
   <tbody class="list">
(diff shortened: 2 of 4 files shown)


andk:
And deployed

PR #588, opened 2026-04-26
Containerize

domm:
This commit-set makes it possible to run PAUSE on your local dev machine using `docker compose`.

See `doc/local-dev-docker-compose.md` for usage instructions.

No part of PAUSE itself was changed, I "just" created a containerized version and fiddled with `docker-compose` to get it working. It is already using `app_2026.psg`!


andk:
Thanks, applied

PR #590, opened 2026-05-09
We should be using Debian Trixie (13), not Debian Bookworm (12).

rspier:


rspier:
Once https://github.com/Perl/docker-perl-tester/pull/176 lands, we can update the GitHub action used for testing.

